PR #89, opened 2017-10-25
Added analog oversampling

Ureloc:
Implemented analog oversampling which allows any analog pin to be automatically sampled muliple times (upto 256) when analogRead() is used. An average of the samples is then returned. Useful to improve noise performance without extra code.

Implemented Uart::flush(). Was there a specific reason for not originally implementing this? Seems like a very useful addition.

### DIFF
--- a/cores/nRF5/Uart.cpp
+++ b/cores/nRF5/Uart.cpp
@@ -186,6 +186,7 @@ void Uart::end()
 
 void Uart::flush()
 {
+	rxBuffer.clear();
 }
 
 void Uart::IrqHandler()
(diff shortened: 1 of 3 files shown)


mchinand:
Is the if/else needed in lines 132-133? It seems lines 138 and 168 take care of setting saadcBurst (but the equivalent if/else would be `if (ulOversampling > 1)...`) 
Should saadcBurst be set to SAADC_CH_CONFIG_BURST_Enabled only if the input parameter is a power of two?
The new variable `saadcOversampling` isn't used.

Ureloc:
Thank you mchinand. I've removed the unnecessary code and tested.

microbuilder:
Thanks for the PR

PR #93, opened 2017-11-06
Allow HardwarePWM::addPin method to return true if _count == MAX_CHAN…

### LINKED ISSUE #92: HardwarePWM::addPin does not ever return true if there are 4 pins already configured
## HardwarePWM.cpp: Line 81
If a HardwarePWM object already has four pins configured, the `VERIFY((_count < MAX_CHANNELS))` statement in HardwarePWM::addPin does not allow the method to return true if the pin was previously configured in that object.

Method should return true if the pin was previously configured, but still return false if trying to add a new 5th pin?

### Context:
I was trying to PWM 2 RGB LEDs (6 pins) and noticed that calling analogWrite() would not work properly after calling it the first time on each of the 6 output pins.

### Code I used to replicate the error:
```
#define L1_RED 7
#define L1_GRN 15
#define L1_BLU 16

#define L2_RED 25
#define L2_GRN 26
#define L2_BLU 27

void setup() {
  analogWrite(L1_RED, 250);
  delay(1000);
  analogWrite(L1_GRN, 250);
  delay(1000);
  analogWrite(L1_BLU, 250);
  delay(1000);
  analogWrite(L2_RED, 250);
  delay(1000);
  analogWrite(L2_GRN, 250);
  delay(1000);
  analogWrite(L2_BLU, 250);
  delay(1000);

  analogWrite(L1_RED, 220);
  delay(1000);
  analogWrite(L1_GRN, 220);
  delay(1000);
  analogWrite(L1_BLU, 220);
  delay(1000);
  analogWrite(L2_RED, 220);
  delay(1000);
  analogWrite(L2_GRN, 220);
  delay(1000);
  analogWrite(L2_BLU, 220);
  delay(1000);

  analogWrite(L1_RED, 10);
  delay(1000);
  analogWrite(L1_GRN, 10);
  delay(1000);
  analogWrite(L1_BLU, 10);
  delay(1000);
  analogWrite(L2_RED, 10);
  delay(1000);
  analogWrite(L2_GRN, 10);
  delay(1000);
  analogWrite(L2_BLU, 10);
}

void loop() {
  // put your main code here, to run repeatedly:
  // ...nothing
}
```

burbankb:
Allow HardwarePWM::addPin method to return true if _count == MAX_CHANNELS, but pin passed was already configured. Return false if trying to add a new pin more than MAX_CHANNELS. fixes #92

### DIFF
--- a/cores/nRF5/HardwarePWM.cpp
+++ b/cores/nRF5/HardwarePWM.cpp
@@ -78,12 +78,18 @@ void HardwarePWM::setClockDiv(uint8_t div)
  */
 bool HardwarePWM::addPin(uint8_t pin)
 {
-  VERIFY( isPinValid(pin) && (_count < MAX_CHANNELS) );
+  VERIFY( isPinValid(pin) && (_count <= MAX_CHANNELS) );
 
   // Check if pin is already configured
   for(uint8_t i=0; i<_count; i++)
   {
     if (_pwm->PSEL.OUT[i] == pin) return true;
+  }
+
+  if ((_count >= MAX_CHANNELS))
+  {
+    //Pin not already configured, but this HardwarePWM is full
+    return false;
   }
 
   pinMode(pin, OUTPUT);


hathach:
thanks, this is indeed a corner case

PR #101, opened 2017-12-02
Fix & upgrade blink sketch

grandinquisitor:
It looks like the blinky sketch had been overwritten with some test code.

Figured might as well upgrade it to blink both red and blue LEDs.


hathach:
thanks, the issue is fixed in a develop branch. We will merge soon enough. Furthermore, I like to stay blinking only RED LED since future nrf52 board may only has one LED (not sure).

PR #102, opened 2017-12-02
Use pip install . for nrfutil & how to fix it if you installed a different version

grandinquisitor:
pip install . is equivalent to and has several advantages and no disadvantages that I can tell vs. python setup.py install

also added instructions on how to install the correct version if you already had one installed.


hathach:
nrfutil is migrated to python3 adafruit-nrfutil. It can be installed with PyPi package. This should be closed, thank you.

```
pip3 install --user adafruit-nrfutil
```

PR #104, opened 2017-12-12
enable BLEHidAdafruit to send key codes larger than 101.

ogatatsu:
Japanese keyboard requires key code larger than 101


hathach:
thanks for PR. Much appreciated :D

PR #122, opened 2018-02-21
BLEHidGeneric: fix missing writeN() function update

hodrob84:
Otherwise getting the following error:
uint16_t BLECharacteristic::write(const void*, uint16_t): 431: verify failed, error = NRF_ERROR_INVALID_ADDR

Signed-off-by: Robert Hodaszi <hodrob84@gmail.com>


hathach:
thanks, we did some renaming and didn't clean up very well :)

Issue #68, opened 2017-07-24
support Servo Libraries 


hathach:
Standard Servo Libraries is already ported to work with nrf52 on Primo board
https://github.com/arduino-libraries/Servo/tree/master/src/nrf52

It is based on several Nordic peripheral driver in SDK which we lack ( due to restricted license). 
https://github.com/arduino-org/arduino-core-nrf52/tree/master/cores/arduino/components

However since the latest sdk 13 open the license to BSD like. We could include those and get the Servo support

hathach:
to many differences, It is easier to implement Servo with our own HardwarePWM :(

microbuilder:
Anything that's compatible at the higher level API level is fine, I think.

hathach:
the hardware PWM works fine, we just need to to wrap it up with the servo API and have some tests to make sure it works. I am finding those little servo to hook up :D

hathach:
the update code should work, though more tests should be carried out.

ichbtm:
Don't know if I post at the right place.
Your implementation use hardcoded pulse limits and doesn't use user values given to the attach function.
More, input of writeMicroseconds is not a µs value.
Maybe the following modifications should be nice ?

Servo.h :
//---convert int8_t to int16_t to be able to store a pulse width
   int16_t min;                       // minimum pulse µs
   int16_t max;                       // maximum pulse µs

servoTimers.h :
//---suppress hardcoded duty cycle values
//#define MIN_PULSE 55
//#define MAX_PULSE 284

//---add some named constants for clarity
//PWM_PRESCALER_PRESCALER_DIV_128 -> NRF_PWM_CLK_125kHz -> resolution 8µs
//MaxValue = 2500 -> PWM period = 20ms
//20ms - 50Hz
#define RESOLUTION 8
#define MAXVALUE 2500
#define CLOCKDIV PWM_PRESCALER_PRESCALER_DIV_128

Servo.h :
//--- use Servo.h limits by default
uint8_t Servo::attach(int pin) {
	return this->attach(pin, MIN_PULSE_WIDTH, MAX_PULSE_WIDTH);
}

//--- use Servo.h limits and store user ones in structure
uint8_t Servo::attach(int pin, int min, int max) {
  if (this->servoIndex < MAX_SERVOS) {
    pinMode(pin, OUTPUT);                                   // set servo pin to output
    servos[this->servoIndex].Pin.nbr = pin;

[...]

PR #134, opened 2018-03-30
fix hardcoded pulse limits

ichbtm:
As notified there https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/68 :
Your implementation use hardcoded pulse limits and doesn't use user's values given to the attach function.
In addition, microseconds functions doesn't use ou return a µs value.

I had to modify the type of min and max parameters to allow storage of values up to 2500µs... maybe it had would be better to add new params (ex : minus and maxus) to avoid confusion with other ARCH...

### DIFF
--- a/libraries/Servo/src/Servo.h
+++ b/libraries/Servo/src/Servo.h
@@ -111,8 +111,8 @@ public:
   bool attached();                   // return true if this servo is attached, otherwise false 
 private:
    uint8_t servoIndex;               // index into the channel data for this servo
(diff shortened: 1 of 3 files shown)


hathach:
@ichbtm thank you very much for your help, somehow this issue fall off our radar until recently :(

hathach:
@ichbtm there is a new push that fix the incorrect freq due to a bug in PWM module. Thank you again

PR #140, opened 2018-04-06
Fix crash caused by incorrectly formatted (e.g. 1 byte) device reports.

swilson:


hathach:
Thanks. it is unlikely error but deadly one

PR #141, opened 2018-04-06
Added battery service client

swilson:


libraries/Bluefruit52Lib/src/clients/BLEClientBas.h
```diff
@@ -0,0 +1,58 @@
...
+
+class BLEClientBas : public BLEClientService
+{
+  public:
+    BLEClientBas(void);
+
+    virtual bool  begin(void);
+    virtual bool  discover(uint16_t conn_handle);
+
+    uint8_t getLevel(void);
```
hathach: maybe simply call it read() which is commonly used

libraries/Bluefruit52Lib/src/clients/BLEClientBas.cpp
```diff
@@ -0,0 +1,82 @@
...
+uint8_t BLEClientBas::getLevel(void)
+{
+  uint8_t level = 0;
+  
+  ble_gattc_handle_range_t bck_range = Bluefruit.Discovery.getHandleRange();
+
+  // Set discovery handle to BAS service
+  Bluefruit.Discovery.setHandleRange(_hdl_range);
+
+  BLEClientCharacteristic chr(UUID16_CHR_BATTERY_LEVEL);
```
hathach: the chr should be part of the service, so that we don't have to perform discovery every time we read the battery value

hathach:
Thanks for the PR. We are appreciated, however before merging this. Would you mind taking a bit more time to check our comments above.

hathach:
close due to lack of user interaction

PR #146, opened 2018-04-19
add method sclLow() for power saving devices - fix printFloat

jaimealemany:
sclLow() in Wire_nRF52 useful to shut down some I2C / smbus devices between end() and start()
fixed a bug in Print.cpp where printFloat won't print (and will fail) if more than 9 digits requested.

### DIFF
--- a/cores/nRF5/Print.cpp
+++ b/cores/nRF5/Print.cpp
@@ -228,10 +228,20 @@ size_t Print::printNumber(unsigned long n, uint8_t base)
 size_t Print::printFloat(double number, uint8_t digits)
 {
   char buf[256];
+  char format[6];
   size_t s=0;
 
-  char format[] = "%.0f";
-  format[2] += digits;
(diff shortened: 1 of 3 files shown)


hathach:
Thanks for spotting the issue with printFloat format and the PR. Sorry for the huge delay. 
For sclLow() is not part of standard Arduino Wire API, so  I hesitate to merge. Can you elaborate it more.

cores/nRF5/Print.cpp
```diff
@@ -228,10 +228,20 @@ size_t Print::printNumber(unsigned long n, uint8_t base)
 size_t Print::printFloat(double number, uint8_t digits)
 {
   char buf[256];
+  char format[6];
   size_t s=0;
 
-  char format[] = "%.0f";
-  format[2] += digits;
+  if (digits < 10)
```
hathach: nice bug catch :)

PR #149, opened 2018-05-03
Fix upload bootloader on Mac OS

stefandz:
Unless I am doing something silly, my symlink to /usr/local/bin/nrfjprog wasn't getting picked up by the code and so bootloader upload was failing (as just nrfjprog was getting called, and I don't think I should be setting symlinks to /bin). If I am getting something very wrong here please let me know so I can fix at my end!


hathach:
nrfutil is migrated to python3 adafruit-nrfutil. Pre-buitin executable for windows and mac are included in BSP. This should be closed, thank you.

PR #158, opened 2018-06-08
Bug Fix: UART RX IRQ

Ureloc:
A bug was identified in the UART IRQ where rapidly receiving data would cause the IRQ to stop responding. The rest of the system would continue to run just fine (ie. no hardfault occurs). The problem stems from clearing nrfUart->EVENTS_RXDRDY at the end of the IRQ rather than at the beginning.

To confirm the existence of the bug do the following:

1. Use the attached .ino. This is simply the hid_keyboard example running at 921,600 bps
2. Connect to any Bluetooth device (I connected with my phone)
3. Send a test sequence at 921,600 bps using any UART terminal. 

The Feather should stop responding after a few characters (+- 20 in my case).

I've attached a session using the YAT UART terminal that includes the test sequence I used (use the "TEST" shortcut button on the right).

These 2 screenshots show the results. Blue = from YAT to Feather, Purple = Echo from Feather. As can be seen the Feather stops sending Echos after 20 characters with old code, but continues to respond indefinitely with new code.

Old Code:
old_code

New Code:
new_code

[UART_Test.zip](https://github.com/adafruit/Adafruit_nRF52_Arduino/files/2083097/UART_Test.zip)


hathach:
Thanks for the detail description and your effort for the fix. We will test it out and include it in the next release.

hathach:
@Ureloc It has been a while. Thank you very much for the fix. Nordic specs does indeed insist that RXDRDY must be cleared before reading RXD. Sorry for a huge delay for this great PR. Since Uart affect lots of user, and we were too busy with nrf52840.

```
To secure that the CPU can detect all incoming RXDRDY events through the RXDRDY event register, the
RXDRDY event register must be cleared before the RXD register is read. The reason for this is that the
UART is allowed to write a new byte to the RXD register, and therefore can also generate a new event,
immediately after the RXD register is read (emptied) by the CPU.
```

PR #169, opened 2018-08-21
Fixed output of dbgDumpMemory when buffer size > 255 bytes

airbornemint:
Output of `dbgDumpMemory` is completely broken when buffer size is > 255 because the code to handle that case multiplies a char (instead of its int value) by 2. This fixes the problem.

Before the fix, the output of dumping a buffer between 2^8 and 2^16 bytes long looks like:

    00:  255lX 255lX 255lX 255lX

because `‘2’ * 2` happens to be `‘d’` so the format string ends up being `“%0dlx”`. Beyond 2^16 bytes, it would fail even harder, as `‘d’ * 2` isn’t a valid format character. I didn’t try it. 

This is an untested code change; I am not set up to build the core and test it here, I just ran into this bug while debugging a library I am working on.


hathach:
thanks for the fix

PR #170, opened 2018-08-26
BLEGATT_ATT_MTU_MAX should be BLE_GATT_ATT_MTU_MAX

airbornemint:
(For consistency with every other BLE_GATT* #define)


hathach:
thanks, you're right. The API around this is still experimental, we didn't name it properly.

PR #177, opened 2018-09-01
Add a way to set connection supervision timeout

airbornemint:
Currently peripheral connection supevision timeout is hardcoded to 2000ms; it should be configurable.


hathach:
thanks, this is advanced timing tuning. Many will find these functions useful

Issue #171, opened 2018-08-29
attachInterrupt vs. SoftDevice


Nenik:
I have run into an issue when using GPIO interrupts.
If I call attachInterrupt() before Bluefruit.begin(), the Bluefruit.begin() fails with NRF_ERROR_SDM_INCORRECT_INTERRUPT_CONFIGURATION, which I suspect might be related to the interrupt priority used by the attachInterrput() implementation.
Calling Bluefruit.begin() first, the interrupr callback works well.
This would really be worth documenting at least.
But should the change in the interrupt priority fix the issue, that would be even better.

hathach:
Ah yeah, it is mentioned multiple times in the forum. Bluefruit.begin() enable sd, which will affect the vector table in some ways. Some irq may not work if called before that. We haven't looked into the details to resolve this, maybe later when we got more free time.

Nenik:
Pull request: https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/179

hathach:
thanks for PR, this should be fixed already.

PR #179, opened 2018-09-01
Fix upstream issue #171: Better choice of pin interrupt priority

Nenik:
Nordic SoftDevice needs to have control over lowest priority interrupts
for reliable, timing sensitive operation. If user code tries to use
such a low priority level, SoftDevice refuses to start.
One way to overcome this limitation is to start the SoftDevice before
attaching pin change interrupts, though in such a case, SD will
cap the effective priority of the user interrupt anyway and it is
a common mistake API users do. This change simply changes the priority
to 2, which is legal (for fast interrupt handlers anyway) under SoftDevice,
no matter if attached before of after SD start.

### DIFF
--- a/cores/nRF5/WInterrupts.c
+++ b/cores/nRF5/WInterrupts.c
@@ -42,7 +42,7 @@ static void __initialize()
 
   NVIC_DisableIRQ(GPIOTE_IRQn);
   NVIC_ClearPendingIRQ(GPIOTE_IRQn);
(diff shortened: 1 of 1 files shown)


hathach:
@Nenik 2 should be the maximum irq just to be safe I think. I post an SD irq model from SD spec here.

https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/171

PR #186, opened 2018-09-22
Update BLEScanner.cpp

Ryscho:
According to this post on the Nordic DevZone Forum (https://devzone.nordicsemi.com/f/nordic-q-a/14643/s130-permanent-advertising-scanning-during-operation), scanning automatically stops after receiving a BLE_GAP_EVT_TIMEOUT{BLE_GAP_TIMEOUT_SRC_SCAN} event. The _runnning [sic] flag should be set to false in this case.


hathach:
thanks alot, sorry for the way, having busy with nrf52840 support until now :D

Issue #183, opened 2018-09-07
MIDI central device notification problem


jardous:
Hello,

I have two Feathers nRF52. One is a peripheral programmed with the examples/blemidi and it works fine when I connect from Linux or iPad.

Unfortunately it seems that setting notification does not seem to work. The BLEClientCharacteristic.enableNotify() does not seem to have any effect and the actual data notification are not sent at all.

Here is sketch I use:
```c++
#include <bluefruit.h>

BLEClientService        midiService(BLEMIDI_UUID_SERVICE);
BLEClientCharacteristic midiChar(BLEMIDI_UUID_CHR_IO);
 
void setup()
{
  Serial.begin(115200);
 
  Serial.println("Bluefruit52 Central Custom HRM Example");
  Serial.println("--------------------------------------\n");
 
  // Initialize Bluefruit with maximum connections as Peripheral = 0, Central = 1
  // SRAM usage required by SoftDevice will increase dramatically with number of connections
  Bluefruit.begin(0, 1);
 
  Bluefruit.setName("Bluefruit52 Central");
 
  // Initialize HRM client
  midiService.begin();
 
  // Initialize client characteristics of HRM.
  // Note: Client Char will be added to the last service that is begin()ed.
  //bslc.begin();
 
  // set up callback for receiving measurement
  midiChar.setNotifyCallback(midi_notify_callback);
  midiChar.begin();
 
  // Increase Blink rate to different from PrPh advertising mode
  Bluefruit.setConnLedInterval(250);
 
  // Callbacks for Central
  Bluefruit.Central.setDisconnectCallback(disconnect_callback);
  Bluefruit.Central.setConnectCallback(connect_callback);
 
  /* Start Central Scanning
   * - Enable auto scan if disconnected
   * - Interval = 100 ms, window = 80 ms
   * - Don't use active scan
   * - Filter only accept HRM service
   * - Start(timeout) with timeout = 0 will scan forever (until connected)
   */
  Bluefruit.Scanner.setRxCallback(scan_callback);
  Bluefruit.Scanner.restartOnDisconnect(true);
  Bluefruit.Scanner.setInterval(160, 80); // in unit of 0.625 ms
  Bluefruit.Scanner.filterUuid(midiService.uuid);
  Bluefruit.Scanner.useActiveScan(false);
  Bluefruit.Scanner.start(0);                   // // 0 = Don't stop scanning after n seconds
}
 
void loop()
{
  // do nothing
}
 
[...]

PR #204, opened 2018-11-21
Adding BLEMidi central support

jardous:
This patch set adds BLEMidi central device support.

At the current state it is not possible to use it with the MIDI library but it is already possible to send and receive data from any BLEMidi peripheral device.


hathach:
@jardous thanks for the PR, this is interesting. Do you have an example sketch there, maybe add it in as well to https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/master/libraries/Bluefruit52Lib/examples/Central

This will help other to get started with Client Midi. We could later make the example compatible with other MIDI library later on

jardous:
This is still a prototype. The BLEMidi packet headers have to be added separately though...

I hope somebody would have an idea on how to create a MIDI instance and use it transparently like the peripheral mode does

hathach:
ah, too bad, I am not really good with MIDI. I got lots of help with prph MIDI implementation. Prph is mostly what people use right ? or am I missing any fun with MIDI central.

jardous:
well - people are not using it, because it is not implemented :)

The problem is that you always have to use you phone as central device. What if you want to transfer your MIDI signal without a phone? Yes, most of the devices a peripherals. But you always need a central one to connect to. Having support for central device enables developers implement also the other side.

I am currently working on wireless MIDI controller for guitar effect processor. The processor has a MIDI input and output so I have to eiter use something like Quicco Sound mi.1 that acts as peripheral and get nRF52 acting as central or I have to implement both.

hathach:
I could help with adding the client MIDI class. I forgot the MIDI specs. but I think we should leave the MIDI parser to external library like how the MIDI prph did. Client MIDI should focus only to get the communication going on with prhp. Can you tell me which function that client MIDI need: read() and write() right ? Though I am still in middle of other things, if you don't mind the wait.

hathach:
Question: the nrf52840 has native usb, we have an plan to add USB Midi class implementation for it, do you think it could help to make thing more fun with the USB ( I am basically clueless about musical stuffs).

jardous:
> I could help with adding the client MIDI class. I forgot the MIDI specs. but I think we should leave the MIDI parser to external library like how the MIDI prph did. Client MIDI should focus only to get the communication going on with prhp. Can you tell me which function that client MIDI need: read() and write() right ? Though I am still in middle of other things, if you don't mind the wait.

It basically needs adding and removing the BLEMidi headers. The protocol is described here: http://www.hangar42.nl/wp-content/uploads/2017/10/BLE-MIDI-spec.pdf

You already have it implemented for a peripheral in services/BLEMidi.c
But the stripping headers could be implemented better I guess. You do not check the message timestamps there for example. That could be a nice improvement.

teafella:
I'm gonna try to get this up and running see if I can get a clock into my modular from the OP-Z in the next couple of days. If you have any examples to share that would be great!

Gonna chime in here as an electronic musician @hathach ; 

USB Midi class compliance would be immensely helpful to me. From my experience, BLEMidi (regardless of latency) simply isn't as reliable as USB at the moment. This is not a problem in the studio but can cause a train wreck at a show.

Compliance would also open up the possibility to easily retrofit exiting midi controllers to BLE and get more people excited about the standard since musicians have a deep distrust for everything wireless.

ladyada:
USB MIDI support has a long chain of dependencies that need to be in place: tinyUSB support, then creating wrappers for MIDI in arduino. will it happen? yes! do we have any ETA whatsoever? nope! :)

jardous:
did anybody try to use both peripheral and central device to connect to each other? I can not get the notification working... Anybody succeeded?

hathach:
@teafella you may want to call requestPairing() before enableNotify() see more detail here https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/183

Thanks for testing it out, I am able to resume to Arduino nrf52 develop now. For better through, there is a bit of tweaking config and fine-tune parameter and several other config to get "the best" . I will try to come up with an example or tutorial on throughput/bandwidth later. There is lots of changes in the latest ble stack regarding this.

hathach:
> > I could help with adding the client MIDI class. I forgot the MIDI specs. but I think we should leave the MIDI parser to external library like how the MIDI prph did. Client MIDI should focus only to get the communication going on with prhp. Can you tell me which function that client MIDI need: read() and write() right ? Though I am still in middle of other things, if you don't mind the wait.
> 
> It basically needs adding and removing the BLEMidi headers. The protocol is described here: http://www.hangar42.nl/wp-content/uploads/2017/10/BLE-MIDI-spec.pdf
> 
> You already have it implemented for a peripheral in services/BLEMidi.c
> But the stripping headers could be implemented better I guess. You do not check the message timestamps there for example. That could be a nice improvement.

For the timestamp, I have no idea though, it is part of the protocol. I really not sure if application want that or not ( I have no clues, never play anything that makes sound in my life :( ). If you think it should be dropped, then comment those line out with brief description to help me digest later when merging. 

Btw, which MIDI library we should use for central role ? We are currently use https://github.com/FortySevenEffects/arduino_midi_library for MIDI prph. I am not sure if we could use it for central ?

teafella:
Not really sure wether timestamps are necessary, my guess that most of the time if you receive MIDI you process it ASAP in any given audio application. There may be some nuances if you are trying to sync midi clocks and compensate for latency? Perhaps a toggle is appropriate based on application.

Also: My Peripheral wasn't being registered unless I was doing an active scan, not sure why but might be worth a comment in the example code if you guys know the reason.

@hathach Would be great to see some throughput optimization examples I'm not educated enough on BLE to grasp it myself.

jardous:
well @teafella, the standard is there so we have to comply to it.
Those timestamps are just increasing number. The current implementation does simply strip it out, but they might be useful to find missing messages, etc.
Of course we can always create a separate characteristic and implement it on our own if we have our both central and peripheral.

Yes, @hathach. I am also looking forward that throughput optimizations. Currently the default throughput is not big enough to transfer full speed MIDI data (31250 baud).

hathach:
> Not really sure wether timestamps are necessary, my guess that most of the time if you receive MIDI you process it ASAP in any given audio application. There may be some nuances if you are trying to sync midi clocks and compensate for latency? Perhaps a toggle is appropriate based on application.

They mention it in the specs, it may be useful for application to analyze it, though currently the example doesn't doing anything with timestamp yet.
 
`In this protocol, MIDI messages are prepended with timestamps, buffered, and accessed via a BLE
characteristic. It is important to note that the frequency with which data can be transmitted
(determined by the connection interval) places a lower bound on latency. Millisecond-resolution
timestamps allow for jitter comparable to the USB Class Specification for MIDI Devices version 1.0.
As Bluetooth technology develops and the usable connection intervals are lowered, latency will
improve correspondingly. However, delays in wireless packet delivery may occur unexpectedly at
any time, temporarily increasing latency and jitter.`

> Also: My Peripheral wasn't being registered unless I was doing an active scan, not sure why but might be worth a comment in the example code if you guys know the reason.

I don't know that either, it could be specific to your prph. central midi is not meant to work with all the prph out there. So you got to make modification to get whatever it works for you. Though I can help analyze it if you could enable debug level and post the output here (in quote).

> well @teafella, the standard is there so we have to comply to it.
> Those timestamps are just increasing number. The current implementation does simply strip it out, but they might be useful to find missing messages, etc.
> Of course we can always create a separate characteristic and implement it on our own if we have our both central and peripheral.

I think we should stick to the specs  :D, it may come in handy later on when you want to sync several devices !?

> @hathach Would be great to see some throughput optimization examples I'm not educated enough on BLE to grasp it myself.
> Yes, @hathach. I am also looking forward that throughput optimizations. Currently the default throughput is not big enough to transfer full speed MIDI data (31250 baud).

There are several factors/settings that will affect the throughput, it may take a bit of time, I am trying to troubleshooting remaining bugs/issues (people have been waiting for some time) and works with other project as well. It is good enough if we could have some proof of concept for now.

PR #229, opened 2019-02-14
Bug Fix: Incorrect cast on BLECharacteristic::read32()

techno:
`read32()` does not work correctly when payload > 0x0000ffff.
Because `read32()` has incorrect cast to`uint16_t` inside the function...


henrygab:
I'm don't have write access, but this is a clear bug, and the fix is similarly clear and simple.  Hope they integrate this ASAP.

hathach:
thank you very much for the PR.

PR #236, opened 2019-03-11
Fix return statements

brijohn:
This fixes several warnings about not returning values when there is an expected return value.

In particular the missing returns in BLEClientHidAdafruit where actually causing an issue when using PlatformIO to compile the central_hid.ino example where calling hid.begin() would cause the board to crash and hang.


jeremypoulter:
:+1: Good work, you will probably see some warning fixes coming from me to ;-)

hathach:
thanks a lot @brijohn , that is quite a bit of warnings I made there. I will need to increase the warning level of Arduino IDE for better caching these.

PR #237, opened 2019-03-11
LittleFS extension rework

jeremypoulter:
Restructured so the class can be extended to add support for external flash chips. This would be used as follows:

ExtFlashFS.h

```C++
#ifndef EXT_FLASH_FS_MANAGER_H
#define EXT_FLASH_FS_MANAGER_H

// Internal Flash uses ARM Little FileSystem
// https://github.com/ARMmbed/littlefs
#include "InternalFS.h"

class ExtLittleFS : public LittleFS
{
  protected:
    virtual int _flash_read (lfs_block_t block, lfs_off_t off, void *buffer, lfs_size_t size);
    virtual int _flash_prog (lfs_block_t block, lfs_off_t off, const void *buffer, lfs_size_t size);
    virtual int _flash_erase (lfs_block_t block);
    virtual void _flash_erase_all();
    virtual int _flash_sync ();

  public:
    ExtLittleFS(void);
    ~ExtLittleFS();

    bool begin(void);
};

extern ExtLittleFS ExtFlashFS;

#endif // !EXT_FLASH_FS_MANAGER_H
```

ExtFlashFS.cpp

```C++
#ifndef ENABLE_DEBUG_FLASH_FS
#undef ENABLE_DEBUG
#endif

#include "ExtFlashFS.h"
#include "MicroDebug.h"
#include "SPIFlash.h"

#ifndef ENABLE_DEBUG_FLASH_FS
#define VERIFY_LFS(...)       _GET_3RD_ARG(__VA_ARGS__, VERIFY_ERR_2ARGS, VERIFY_ERR_1ARGS)(__VA_ARGS__, NULL)
#else
#define VERIFY_LFS(...)       _GET_3RD_ARG(__VA_ARGS__, VERIFY_ERR_2ARGS, VERIFY_ERR_1ARGS)(__VA_ARGS__, dbg_strerr_lfs)
#endif

// External SPI Flash uses ARM Little FileSystem
// 
//    | | |     .---._____
//   .-----.   |          |
// --|o    |---| littlefs |
// --|     |---|          |
//   '-----'   '----------'
//    | | |
// 
// https://github.com/ARMmbed/littlefs

//! Default Flash Chip
//! CYPRESS
#define CYPRESS_FLASH_JEDEC_ID  0x0160 

//! ISSI Flash
//! jedec_MANUF_ID = 0x9D
//! jedec_MEM_TYPE = 0x60
//! jedec_CAPACITY = 0x19
#define ISSI_FLASH_JEDEC_ID     0x9D60

#define FLASH_BLOCK_SIZE      0x1000 // 4 kByte block size

#define LFS_FLASH_ADDR        0x00000
#define LFS_FLASH_TOTAL_SIZE  0x2000000
#define LFS_BLOCK_SIZE        FLASH_BLOCK_SIZE 
#define LFS_READ_SIZE         (LFS_BLOCK_SIZE / 16) 
#define LFS_PROG_SIZE         (LFS_READ_SIZE * 1)

static SPIFlash flash(SS, CYPRESS_FLASH_JEDEC_ID); // 0x0160 is Cypress Flash JEDEC code

ExtLittleFS ExtFlashFS;

static inline uint32_t lba2addr(uint32_t block)
{
  return ((uint32_t) LFS_FLASH_ADDR) + block * LFS_BLOCK_SIZE;
}

static inline void flashWait() {
  while(flash.busy()){
    delay(10);
  }
}

ExtLittleFS::ExtLittleFS(void) : 
  LittleFS( LFS_READ_SIZE, LFS_PROG_SIZE, LFS_BLOCK_SIZE, LFS_FLASH_TOTAL_SIZE / LFS_BLOCK_SIZE, 128)
{
}

ExtLittleFS::~ExtLittleFS()
{
}

bool ExtLittleFS::begin (void)
{
  if ( _begun ) return true;

  //! Initialize SPI
  DBUGLN("Initialising SPI bus");
  SPI.begin();

  //! Initialize flash and Check for Cycpress Flash chip ID
  DBUGLN("Initialising flash");
  if (true == flash.initialize()) 
  {
    // Serial.println("Init Flash OK!");
  }
  else
  {
    //! Check for ISSI Flash ID
    uint16_t jedecID = flash.readDeviceId();
    if ( (0 == jedecID) || (ISSI_FLASH_JEDEC_ID == jedecID))
    {
      // Serial.println("Init Flash OK!");
    }
    else
    {    
      Serial.print("JedecID = ");
      Serial.println(jedecID);

      Serial.println("Init Flash FAIL!");
    }
  }

  //===========================================================================
  // delay 30ms to allow Flash to startup
  // This is required because the BleManager needs to read the flash via
  // the deviceManager for the BLE DeviceName
  //===========================================================================
  flashWait();

  return LittleFS::begin();
}

int ExtLittleFS::_flash_read (lfs_block_t block, lfs_off_t off, void *buffer, lfs_size_t size)
{
  DBUGF("_flash_read block %d, off %d, buffer %p, size %d", block, off, buffer, size);

  uint32_t addr = lba2addr(block) + off;
  flash.readBytes(addr, buffer, size);
  flashWait();

  return 0;
}

// Program a region in a block. The block must have previously
// been erased. Negative error codes are propogated to the user.
// May return LFS_ERR_CORRUPT if the block should be considered bad.
int ExtLittleFS::_flash_prog (lfs_block_t block, lfs_off_t off, const void *buffer,
                         lfs_size_t size)
{
  DBUGF("_flash_prog block %d, off %d, buffer %p, size %d", block, off, buffer, size);

  uint32_t addr = lba2addr(block) + off;
  flash.writeBytes(addr, buffer, size);
  flashWait();

  return 0;
}

// Erase a block. A block must be erased before being programmed.
// The state of an erased block is undefined. Negative error codes
// are propogated to the user.
// May return LFS_ERR_CORRUPT if the block should be considered bad.
int ExtLittleFS::_flash_erase (lfs_block_t block)
{
  DBUGF("_flash_erase block %d", block);

  uint32_t addr = lba2addr(block);
  flash.blockErase4K(addr);
  flashWait();

  return 0;
}

void ExtLittleFS::_flash_erase_all()
{
  DBUGF("_flash_erase_all");

  for ( uint32_t addr = LFS_FLASH_ADDR; addr < LFS_FLASH_ADDR + LFS_FLASH_TOTAL_SIZE; addr += FLASH_BLOCK_SIZE )
  {
    flash.blockErase4K(addr);
  }
}

// Sync the state of the underlying block device. Negative error codes
// are propogated to the user.
int ExtLittleFS::_flash_sync()
{
  DBUGF("_flash_sync");

  return 0;
}
```

### DIFF
--- a/libraries/FileSystem/src/InternalFS.cpp
+++ b/libraries/FileSystem/src/InternalFS.cpp
@@ -118,79 +118,30 @@ static inline uint32_t lba2addr(uint32_t block)
   return ((uint32_t) LFS_FLASH_ADDR) + block * LFS_BLOCK_SIZE;
 }
 
-static int _iflash_read (const struct lfs_config *c, lfs_block_t block, lfs_off_t off, void *buffer, lfs_size_t size)
-{
-  (void) c;
-
-  uint32_t addr = lba2addr(block) + off;
-  flash_nrf5x_read(buffer, addr, size);
-
-  return 0;
-}
-
-// Program a region in a block. The block must have previously
-// been erased. Negative error codes are propogated to the user.
-// May return LFS_ERR_CORRUPT if the block should be considered bad.
-static int _iflash_prog (const struct lfs_config *c, lfs_block_t block, lfs_off_t off, const void *buffer,
-                         lfs_size_t size)
-{
-  (void) c;
-
-  uint32_t addr = lba2addr(block) + off;
-  flash_nrf5x_write(addr, buffer, size);
-
-  return 0;
-}
-
-// Erase a block. A block must be erased before being programmed.
-// The state of an erased block is undefined. Negative error codes
-// are propogated to the user.
-// May return LFS_ERR_CORRUPT if the block should be considered bad.
-static int _iflash_erase (const struct lfs_config *c, lfs_block_t block)
-{
-  (void) c;
-
-  uint32_t addr = lba2addr(block);
-
-  // implement as write 0xff to whole block address
-  for(int i=0; i <LFS_BLOCK_SIZE; i++)
-  {
-    flash_nrf5x_write8(addr + i, 0xFF);
-  }
-
-  // flash_nrf5x_flush();
-
-  return 0;
-}
-
-// Sync the state of the underlying block device. Negative error codes
-// are propogated to the user.
-static int _iflash_sync (const struct lfs_config *c)
-{
-  (void) c;
-  flash_nrf5x_flush();
-  return 0;
-}
-
 //--------------------------------------------------------------------+
 // Implementation
 //--------------------------------------------------------------------+
 LittleFS InternalFS;
 
-LittleFS::LittleFS (void)
+LittleFS::LittleFS (void) :
+  LittleFS( LFS_BLOCK_SIZE, LFS_BLOCK_SIZE, LFS_BLOCK_SIZE, LFS_FLASH_TOTAL_SIZE / LFS_BLOCK_SIZE, 128)
+{
+}
+
+LittleFS::LittleFS (lfs_size_t read_size, lfs_size_t prog_size, lfs_size_t block_size, lfs_size_t block_count, lfs_size_t lookahead)
 {
   varclr(&_lfs_cfg);
-  _lfs_cfg.context = NULL;
+  _lfs_cfg.context = this;
   _lfs_cfg.read = _iflash_read;
   _lfs_cfg.prog = _iflash_prog;
   _lfs_cfg.erase = _iflash_erase;
   _lfs_cfg.sync = _iflash_sync;
 
-  _lfs_cfg.read_size = LFS_BLOCK_SIZE;
-  _lfs_cfg.prog_size = LFS_BLOCK_SIZE;
-  _lfs_cfg.block_size = LFS_BLOCK_SIZE;
-  _lfs_cfg.block_count = LFS_FLASH_TOTAL_SIZE / LFS_BLOCK_SIZE;
-  _lfs_cfg.lookahead = 128;
+  _lfs_cfg.read_size = read_size;
+  _lfs_cfg.prog_size = prog_size;
+  _lfs_cfg.block_size = block_size;
+  _lfs_cfg.block_count = block_count;
+  _lfs_cfg.lookahead = lookahead;
 
   _begun = false;
   _mounted = false;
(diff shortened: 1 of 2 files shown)


hathach:
Sorry for late response. I have been busy with stm32f port for tinyusb stack recently.  Thanks @jeremypoulter for advance work to extend the littleFS. This look good to me, I could merge it now. Though to be honest, I don't feel right to inherit ExtLittleFS:LittleFS as it is. Ideally we should have LittleFS as abstract class that require `_flash_API`  to be implemented by subclass `InternalLittleFS and ExternalLittleFS`. What do you think about this approach, don't worry, you don't have to change to that now. I could do that later on after the merge. Just want to know your thought.

jeremypoulter:
Yes, that is probably the 'proper' way to do it, no objections from me.

hathach:
as discussed previously, current develop branch separate Little FS and Internal Flash to make it easier for using it with other flash storage such as SPI flash
https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/develop/libraries
- Adafruit_LittleFS
- InternalFileSytem

For SPI flash, you can just change the lfs_config (which is now an input parameter), to perform flash api accordingly. @jeremypoulter let's me know if it causes you any issues. Sorry for breaking changes but it is better to do it sooner than later.

PS: ExternalFS is removed in prefer of 
- https://github.com/adafruit/Adafruit_SPIFlash
- https://github.com/adafruit/Adafruit_QSPI ( just got ported to nrf5x)

sslupsky:
Hi @jeremypoulter I saw your work here regarding the external flash.  Did you do any further work related to this?  Specifically, did you generate and of the flash primitives for any of the SPI nand flash chips available?  I am working with a Toshiba part (TC58CVG2S0HRAIG) and was wondering if there might be something to start with.

@hathach I came across your work reading some posts over on the arm forum.  Thank you for taking the time to do this work.  Do you see any potential issues with using this with the SAMD core?

hathach:
@sslupsky LFS is written in more abstract way here, you could take a look at it. https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/master/libraries/Adafruit_LittleFS

LFS is very portable, it probably should work on any other cores.

sslupsky:
@hathach I have has some success working with littlefs on the SAMD architecture.  Since Adafruit has some SAMD based products,  I am wondering if the Adafruit_LittleFS library could be moved to a general library instead of being part of the nRF52 core library?  That way it could be used by the SAMD and nRF52 without having to maintain two different repositories?

sslupsky:
@hathach I've run into some issues because the Adafruit_LittleFS library has some dependencies on the TinyUSB core.  Seems odd that this should be the case.  It appears to me the dependencies are a set of defines and macros used for TinyUSB.  Perhaps the LittleFS library should have its own versions of these files?

hathach:
@sslupsky This PR is merged, there is no point to unrelated discussion here. Please open a new issue for your feature request

PR #238, opened 2019-03-11
Unmount the FS before formatting

### LINKED ISSUE #235: LittleFS format memory leak
`lfs_format` internally calls `lfs_init` so if you call `LittleFS.format` after `LittleFS.begin` (as would be expected in the Arduino world) it will reallocate the leak the buffers allocated as a result of calling `LittleFS.begin`. 

Is this expected behaviour or a bug.

Note: I have resolved this locally by doing the following;
```c++
bool LittleFS::format (bool eraseall)
{
  if ( eraseall ) {
    _flash_erase_all();
  }
  if(_mounted) {
    VERIFY_LFS(lfs_unmount(&_lfs), false);
  }
  VERIFY_LFS(lfs_format(&_lfs, &_lfs_cfg), false);
  VERIFY_LFS(lfs_mount(&_lfs, &_lfs_cfg), false);

  _mounted = true;

  return true;
}
```

jeremypoulter:
Fixes #235

### DIFF
--- a/libraries/FileSystem/src/InternalFS.cpp
+++ b/libraries/FileSystem/src/InternalFS.cpp
@@ -193,6 +193,7 @@ LittleFS::LittleFS (void)
   _lfs_cfg.lookahead = 128;
 
   _begun = false;
+  _mounted = false;
 }
 
 LittleFS::~LittleFS ()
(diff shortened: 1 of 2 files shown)


hathach:
thank you very much for the fix

PR #239, opened 2019-03-11
Allow the task stack sizes to be configured

jeremypoulter:
(and some other CFG_ items that were in the same files)

This is because the default memory allocations are quite conservative assuming that folks will do lots of work in the callbacks etc (which is fairly good assumption). However this can lead to a lot of 'wasted' RAM. The `dbgMemInfo` tool (which is great BTW) can then be used to look at the memory usage and the values can be fine tuned (at least with PlatformIO).

### DIFF
--- a/cores/nRF5/utility/AdaCallback.h
+++ b/cores/nRF5/utility/AdaCallback.h
@@ -38,9 +38,18 @@
 
 #include "common_inc.h"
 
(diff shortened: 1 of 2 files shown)


hathach:
Ah thanks, right, ideally all CFG_* should be over-writable by make command. But I didn't try to get it work with many platforms to begin with :D

PR #248, opened 2019-04-08
loopback functionality for ancs

ferbass:
# Propose
- ANCS Loopback will create a mirror characteristics of ANCS that can be use to connect to a non-hardware project and exchange the ANCS data through this characteristics. That way the non-hardware project like Apps can handle the raw ANCS that on their side and request the info that they need just sending the request to the ANCS using the Loopback

### DIFF
--- a/libraries/Bluefruit52Lib/examples/Peripheral/ancs_loopback/ancs_loopback.ino
+++ b/libraries/Bluefruit52Lib/examples/Peripheral/ancs_loopback/ancs_loopback.ino
@@ -1,0 +1,185 @@
+#include <bluefruit.h>
+
+BLEDis bledis;
+BLEClientDis bleClientDis;
+BLEAncsLoopback bleancs;
+
+char buffer[128];
+
+const uint8_t ANCS_LOOPBACK_GATT_SERVICE[] =
+{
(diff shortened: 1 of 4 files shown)


hathach:
@ferbass  thank you for your pending PR. I haven't been looked at ancs for some time. Could you give me some info regarding the ANCS Loop back functionality ?

ferbass:
Hello @hathach the loopback functionally is basically a mirror of ANCS. 
The apps cannot communicate directly to the ANCS so a solution is create this loopback to capture the notification in the Application side.
In my case, my app will handle not type of notification like Messages for example but I don't wanna take actions for calls. So with this loopback I can decide what I will do based in the Category of the notification and request the information by the loopback characteristics. Also with that I don't need to do any software update in the hardware side if Apple decide to add a new Category, event, command or etc we just need to do the App Update to handle the information we need.

ferbass:
This is how we use this
loopback

ferbass:
@hathach I'll review my code, add the .ino sample project doing the loopback and send for your review.

hathach:
@ferbass thank you very much for putting lots of effort into this PR. But I am currently in the middle of usb thing, I will review it a bit later. Hope you don't mind

ferbass:
@hathach that's ok, take your time, that's not something urgent for the library, thank you

PR #251, opened 2019-04-13
Fix #250 -- wrong values for g_ADigitalPinMap

### LINKED ISSUE #250: Wrong values in g_ADigitalPinMap?
It appears that five pins are listed twice in the pin map for the Feather nRF52840  Express:
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/b6a8566a531c11bc21b9404615b1a6c836e7ddef/variants/feather_nrf52840_express/variant.cpp#L34-L36
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/b6a8566a531c11bc21b9404615b1a6c836e7ddef/variants/feather_nrf52840_express/variant.cpp#L38
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/b6a8566a531c11bc21b9404615b1a6c836e7ddef/variants/feather_nrf52840_express/variant.cpp#L44

These pins are listed a second time at their own index:
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/b6a8566a531c11bc21b9404615b1a6c836e7ddef/variants/feather_nrf52840_express/variant.cpp#L70-L71

Is this intentional?  This seems like it may be a bug....

henrygab:
This fixes the issues identified in #250.

* Update definition of `PINS_COUNT` and `NUM_DIGITAL_PINS` to reflect mappable pin count for this board (34).
* Update `g_ADigitalPinMap` to define only 34 pins.
* Add comments to `g_ADigitalPinMap`, describing the non-mapped 14 pins, and why they are not listed.

### DIFF
--- a/variants/feather_nrf52840_express/variant.cpp
+++ b/variants/feather_nrf52840_express/variant.cpp
@@ -28,47 +28,74 @@
 
 const uint32_t g_ADigitalPinMap[] =
 {
-  25,  // D0 is UART TX on P0.25
-  24,  // D1 is UART RX on P0.24 
-  10,  // D2 is on P0.10
-  47,  // D3 is LED1 on P1.15
-  42,  // D4 is LED2 on P1.10
-  40,  // D5 on P1.08
-  7,   // D6 on P0.07 
-  34,  // D7 is Button on P1.02
-  16,  // D8 is NeoPixel on P0.16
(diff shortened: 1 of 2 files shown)


hathach:
Thank you very much, look great.

PR #254, opened 2019-04-26
Update adc_vbat.ino to also work with nrf52840

pyro9:
The nrf52840 uses 150K + 150K as the VBATT voltage divider and connects to pin A6 rather than A7.
The cleanest way to support old and new boards is to use an ifdef and change mvToPercent to deal in real mV since full charge isn't 3V on the ADC pin for nrf52840.

I'm not fully satisfied with the new mvToPercent since it looks like the battery never quite reaches 4200 mV.


hathach:
thanks for PR, indeed, we didn't update example to match the vbat pin changes on feather nrf52840. We have added an PIN_VBAT to variant.h to make it easier
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/variants/feather_nrf52840_express/variant.h#L90

PR #260, opened 2019-05-03
Extend SoftwareTimer with option to make it non-repeating, add reset function & ISR-safe functions.

MacGyverNL:
The SoftwareTimer provided by the library is missing a function for resetting it.

Also, I want to reset this timer from an ISR attached by attachInterrupt. Based on the documentation at https://www.freertos.org/FreeRTOS-timers-xTimerResetFromISR.html, I figured it was necessary to use the ISR-safe functions, so there's ISR-safe equivalents included. I'm not quite certain about their name, though; I used ISR because that was what was in my head at the time but it might be more appropriate to name them more in line with Arduino-style "attachInterrupt" thus becoming "resetFromInterrupt" etc.

Finally, the timer I want must be non-repeating, only getting started when an interrupt tells it to start. So I added a begin()-function that adds a boolean to its signature to specify whether the timer repeats if it ticks over or suspends.

See
https://github.com/MacGyverNL/bluepedal/blob/02213ee9dcc495ccdc47f572892f7388aac09a55/bluepedal/bluepedal.ino#L190
and
https://github.com/MacGyverNL/bluepedal/blob/02213ee9dcc495ccdc47f572892f7388aac09a55/bluepedal/bluepedal.ino#L281
for context of how I'm using them to implement a loopless HID.

### DIFF
--- a/cores/nRF5/utility/SoftwareTimer.h
+++ b/cores/nRF5/utility/SoftwareTimer.h
@@ -52,6 +52,11 @@ class SoftwareTimer
       _handle = xTimerCreate(NULL, ms2tick(ms), true, NULL, callback);
     }
 
+    void begin(uint32_t ms, TimerCallbackFunction_t callback, bool repeating)
+    {
+      _handle = xTimerCreate(NULL, ms2tick(ms), repeating, NULL, callback);
+    }
+
     TimerHandle_t getHandle(void)
     {
       return _handle;
@@ -59,6 +64,17 @@ class SoftwareTimer
 
     void start(void) { xTimerStart(_handle, 0); }
     void stop (void) { xTimerStop (_handle, 0); }
+    void reset(void) { xTimerReset(_handle, 0); }
+
+    BaseType_t startFromISR(BaseType_t* pxHigherPriorityTaskWoken) {
+      return xTimerStartFromISR(_handle, pxHigherPriorityTaskWoken);
+    }
+    BaseType_t stopFromISR(BaseType_t* pxHigherPriorityTaskWoken) {
+      return xTimerStopFromISR (_handle, pxHigherPriorityTaskWoken);
+    }
+    BaseType_t resetFromISR(BaseType_t* pxHigherPriorityTaskWoken) {
+      return xTimerResetFromISR(_handle, pxHigherPriorityTaskWoken);
+    }
 
     void setPeriod(uint32_t ms)
     {
--- a/cores/nRF5/utility/SoftwareTimer.h
+++ b/cores/nRF5/utility/SoftwareTimer.h
@@ -47,12 +47,7 @@ class SoftwareTimer
     SoftwareTimer()          { _handle = NULL; }
     virtual ~SoftwareTimer() { if(_handle != NULL) xTimerDelete(_handle, 0); }
 
-    void begin(uint32_t ms, TimerCallbackFunction_t callback)
-    {
-      _handle = xTimerCreate(NULL, ms2tick(ms), true, NULL, callback);
-    }
-
-    void begin(uint32_t ms, TimerCallbackFunction_t callback, bool repeating)
+    void begin(uint32_t ms, TimerCallbackFunction_t callback, bool repeating = true)
     {
       _handle = xTimerCreate(NULL, ms2tick(ms), repeating, NULL, callback);
     }
--- a/cores/nRF5/utility/SoftwareTimer.h
+++ b/cores/nRF5/utility/SoftwareTimer.h
@@ -61,14 +61,25 @@ class SoftwareTimer
     void stop (void) { xTimerStop (_handle, 0); }
     void reset(void) { xTimerReset(_handle, 0); }
 
-    BaseType_t startFromISR(BaseType_t* pxHigherPriorityTaskWoken = NULL) {
-      return xTimerStartFromISR(_handle, pxHigherPriorityTaskWoken);
+    bool startFromISR(void) {
+      BaseType_t ret, xHigherPriorityTaskWoken = pdFALSE;
+      ret = xTimerStartFromISR(_handle, &xHigherPriorityTaskWoken);
+      portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
+      return (ret == pdPASS);
     }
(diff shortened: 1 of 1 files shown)


cores/nRF5/utility/SoftwareTimer.h
```diff
@@ -52,13 +52,29 @@ class SoftwareTimer
...
     TimerHandle_t getHandle(void)
     {
       return _handle;
     }
 
     void start(void) { xTimerStart(_handle, 0); }
     void stop (void) { xTimerStop (_handle, 0); }
+    void reset(void) { xTimerReset(_handle, 0); }
+
+    BaseType_t startFromISR(BaseType_t* pxHigherPriorityTaskWoken) {
```
hathach: `pxHigherPriorityTaskWoken` is often not used, we can have it as default = NULL

cores/nRF5/utility/SoftwareTimer.h
```diff
@@ -52,13 +52,29 @@ class SoftwareTimer
       _handle = xTimerCreate(NULL, ms2tick(ms), true, NULL, callback);
     }
 
+    void begin(uint32_t ms, TimerCallbackFunction_t callback, bool repeating)
```
hathach: I think we should either 
- have repeating as default = true ( 1 begin API)
- have the above begin call this begin(ms, callback, true).

I am fine with either of them

hathach:
Also if you are using attachInterrupt() but didn't need spot-on time, you can have the attachInterrupt() defer callback using ISR_DEFRRED option.

cores/nRF5/utility/SoftwareTimer.h
MacGyverNL: The freeRTOS documentation says that if this function results in the higher priority task being woken, a context switch should be requested before the ISR exits. I'm not certain *why* this is, but it seems unwise to me to not use it. Are you certain ignoring this is safe?

cores/nRF5/utility/SoftwareTimer.h
MacGyverNL: I've implemented the first suggestion.

MacGyverNL:
> Also if you are using attachInterrupt() but didn't need spot-on time, you can have the attachInterrupt() defer callback using ISR_DEFRRED option.

I'm aware of the option, thanks. Most relevant for me is the behaviour of ISR_DEFERRED when multiple interrupts happen before the ISR is run, however -- does it run trhe ISR once for every single one or are all the interrupts rolled into a single run?

Regardless, assuming at some point someone will need spot-on timing for the ISR, and still need to also fire up a timer based on it, I think these additions should be made.

cores/nRF5/utility/SoftwareTimer.h
MacGyverNL: So, from the documentation at https://www.freertos.org/FreeRTOS-timers-xTimerResetFromISR.html:
> If calling xTimerResetFromISR() causes the timer service/daemon task to leave the Blocked state, and the timer service/ daemon task has a priority equal to or greater than the currently executing task (the task that was interrupted), then *pxHigherPriorityTaskWoken will get set to pdTRUE internally within the xTimerResetFromISR() function. If xTimerResetFromISR() sets this value to pdTRUE then a context switch should be performed before the interrupt exits.

And combined with http://www.openrtos.net/FreeRTOS_Support_Forum_Archive/May_2009/freertos_What_if_portYIELD_FROM_ISR_not_used_3282756.html:
> If the interrupt service routine does not call portYIELD_FROM_ISR then the task switch will be delayed until the next time the scheduler is activated, either by another interrupt (including the timer tick), or the current task blocks or yields.

It seems to me that not using this is a terrible idea but worst case will only delay the timer task from actually processing the command to start/stop/reset until the next tick / scheduler invocation.  I still think this is a bad idea but if you still want the API to provide for it I'll add it.

MacGyverNL: Thinking about this some more, I'm not certain whether this "worst case" is actually as mundane as I make it out. What happens if the task that is being interrupted is the IDLE task? What happens if the interrupt happened during device sleep / waitForEvents?

However, it may not be necessary to answer those questions. portYIELD_FROM_ISR, by way of portYIELD, only sets up the scheduler to run immediately after the ISR exits, right? It does *not* interrupt the ISR with a run of the scheduler...

Which means that functionally it doesn't matter how many times in an ISR you would actually call portYIELD; it's just some wasted instructions for actually running portYIELD itself. So what we *could* do is abstract the YIELD logic away from the user -- setup xHigherPriorityTaskWoken in the *fromISR() function, act on it in that function, *then* return the success status of xTimerStartFromISR.

E.g.
```
bool startFromISR(void) {
  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
  bool ret;
  if (xTimerStartFromISR(_handle, &xHigherPriorityTaskWoken) != pdPASS) {
    ret = false;
  } else {
    ret = true;
  }
  portYIELD_FROM_ISR(xHigherPriorityTaskWoken);
  return ret;
```

This shields the library's user from the BaseType_t type, and it ensures a context switch if e.g. the IDLE task was interrupted.

However, this may be a heavier function call than we would like to see inside an ISR.

hathach: @MacGyverNL thanks for in-depth explanation, since this is nrf5x core repo, we don't really worry about other platform. If you tested the above example work, then just remove the parameter and call the portYield() within the startFromISR() giving it a cleaner API.

hathach: @MacGyverNL If you could change the startFromISR() as your above snippet that handle waken task internally, I will merge the PR. We could open another PR/issue if it cause one.

MacGyverNL: I haven't gotten around to testing whether firing it off multiple times from the same interrupt causes an issue.

hathach: judging from freeRTOS it shouldn't be, though, take your time, it isn't urgent at all :)

MacGyverNL: Tested, seems to work.

hathach:
thanks for the PR and follow up :)

Issue #259, opened 2019-04-30
Unable to use GPIO P1.* interrupts on nRF52840


MacGyverNL:
I'm using the Feather nRF52840 revision D, with nRF52 version 0.10.1. I've been toying with the interrupts, and noticed that attaching an interrupt to digital pin 7 (the USR / DFU button) doesn't result in interrupts firing, but running my finger along the bottom of the board *did* trigger interrupts. So I decided to put the entire board in INPUT_PULLUP mode and start grounding interrupted pins, to figure out what was happening.

This has resulted in the following list of pins for which interrupts do not work as expected:
```
02 -> DOES NOT WORK ON D2 (P0.10)
03 -> DOES NOT WORK ON LED1 / D3 (P1.15)
04 -> DOES NOT WORK ON LED2 / D4 (P1.10)
05 -> DOES NOT WORK ON D5 / SEEMS TO SOMETIMES TRIGGER FROM D12 (P1.08, triggers on P0.08)
07 -> DOES NOT WORK ON USRBTN / D7 (P1.02)
13 -> DOES NOT WORK ON D13 (P1.09)
20 -> DOES NOT WORK ON BAT / A6 (P0.29)
```

I have not tried beyond pin 26, because that's the flash and after 33 doesn't seem to be available anywhere on the board either. Everything else, including P8 (the neopixel) works as expected.

Aside from D2 and BAT, the pins that do not allow interrupts are all the pins on P1 that are mapped. I do not have a working theory about D2 and BAT, but the fact that shorting D12 to ground sometimes triggers the interrupt set for D5 is not a coincidence: D5 is P1.08 and D12 is P0.08. 

Since D12 itself works fine, and does not seem to trigger at all on D5, my theory is that there is an issue in how interrupts are attached in cores/nRF5/WInterrupts.c. Unfortunately, I don't understand this chip well enough to know what the issue is, but I did find this topic on Nordic's forums discussing the use of NRF_P0 and NRF_P1 instead of NRF_GPIO:
https://devzone.nordicsemi.com/f/nordic-q-a/41855/nrf52840-gpio-p1-xx-configuration-problem

This may also be related to #250 and #251; however, I put the changes from #251 in place manually, and that did not change anything -- though it may be the case that this would require me to recompile stuff that I'm currently unaware of.

Test code for testing 4 interrupts at a time:

```
int button = 0;
int interrupted1, interrupted2, interrupted3, interrupted4;

void setup() {
[...]

PR #261, opened 2019-05-03
Fix interrupts for device with multiple I/O Ports

MacGyverNL:
Fixes adafruit/Adafruit_nRF52_Arduino#259


hathach:
@MacGyverNL  thank you very much for the fix :+1:

PR #267, opened 2019-05-11
Fix connection handle in BLEHidAdafruit single connection api

ogatatsu:


hathach:
thank you very much, this is such a stupid mistake ~.~

PR #268, opened 2019-05-11
Fix auto-start of advertising when central is connected

ogatatsu:


hathach:
thanks, forgot to change this when changing the API

PR #297, opened 2019-07-03
Corrected location of consumer usage codes

stefandz:
Consumer usage codes are in Adafruit_nRF52_Arduino/cores/nRF5/Adafruit_TinyUSB_Core/tinyusb/src/class/hid/hid.h (or at least that is where I found them) - not BLEHidGeneric.h


hathach:
ah right, thanks, it used to be in BLEHidGeneric.h when usb hid is not added. Move it around recently without updating these comment

Issue #245, opened 2019-03-26
Please add example how to use BLECharacteristic.setReadAuthorizeCallback()


DiazGopar:
Please add example how to use BLECharacteristic.setReadAuthorizeCallback()

hathach:
we don't have time to write example for all APIs, but will be glad to merge it if anyone submit a PR.

PR #300, opened 2019-07-06
Added response to read authorize requests

nsamala:
If you add a `setReadAuthorizeCallback` to a BLECharacteristic, it hangs because there is no response. This PR adds a response to read authorize requests.

### DIFF
--- a/cores/nRF5/nordic/softdevice/s132_nrf52_6.1.1_API/include/ble_gap.h
+++ b/cores/nRF5/nordic/softdevice/s132_nrf52_6.1.1_API/include/ble_gap.h
@@ -594,6 +594,9 @@ enum BLE_GAP_TX_POWER_ROLES
 /**@brief Maximum amount of addresses in the whitelist. */
 #define BLE_GAP_WHITELIST_ADDR_MAX_COUNT (8)
 
+/**@brief Maximum amount of IRKs in a whitelist. */
+#define BLE_GAP_WHITELIST_IRK_MAX_COUNT (8)
+
 
 /**@brief Maximum amount of identities in the device identities list. */
 #define BLE_GAP_DEVICE_IDENTITIES_MAX_COUNT (8)
@@ -693,7 +696,6 @@ typedef struct
   uint8_t addr[BLE_GAP_ADDR_LEN]; /**< 48-bit address, LSB format.
                                        @ref addr is not used if @ref addr_type is @ref BLE_GAP_ADDR_TYPE_ANONYMOUS. */
 } ble_gap_addr_t;
-
 
 /**@brief GAP connection parameters.
  *
--- a/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
+++ b/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
@@ -396,6 +396,18 @@ bool BLEAdvertising::stop(void)
   return true;
 }
 
+bool BLEAdvertising::setWhitelist(ble_gap_addr_t const * const * addrs, uint8_t len) {
+  VERIFY_STATUS( sd_ble_gap_whitelist_set(addrs, len) );
+
+  return true;
+}
+
+bool BLEAdvertising::clearWhitelist() {
+  VERIFY_STATUS( sd_ble_gap_whitelist_set(NULL, 0) );
+
+  return true;
+}
+
 
 void BLEAdvertising::_eventHandler(ble_evt_t* evt)
 {
--- a/libraries/Bluefruit52Lib/src/BLEAdvertising.h
+++ b/libraries/Bluefruit52Lib/src/BLEAdvertising.h
@@ -129,6 +129,9 @@ public:
   void setInterval  (uint16_t fast, uint16_t slow);
   void setIntervalMS(uint16_t fast, uint16_t slow);
 
+  bool setWhitelist(ble_gap_addr_t const * const * addrs, uint8_t len);
+  bool clearWhitelist();
+
   uint16_t getInterval(void);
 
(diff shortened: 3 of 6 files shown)


hathach:
function installed by setReadAuthorizeCallback & setWriteAuthorizeCallback are meant to call `sd_ble_gatts_rw_authorize_reply` with appropriate parameter to response to the request. The reply parameter can be complicated than a simple accept or deny.

ptyork:
Yes, this is complicated, which is why an example would be quite nice per #245. AFAIK, the only way to provide an updated value for a characteristic "on demand" is by responding to a read authorize request. There's no other "I want to read" callback. Which means this is kinda vital to either explain and/or "wrap" in a more elegant way.

Which actually leads to a question I've got. What is the purpose of this library? Is it just to put an "Adafruit" face on top of the nrf SDK? Or is it to provide a layer of abstraction such that you can later decouple this from Nordic? Obviously if the latter, then you'd definitely want to provide some kind of wrapper around the read and write authorize response process. Even if it is more robust than this particular pull request.

Anyway, referring back to #245, I'll be glad to submit a pull request that mods custom_hrm to setProperties(CHR_PROPS_READ|CHR_PROPS_NOTIFY) and implement the callback to provide a heart rate when Notify is disabled. Callback would be pretty ugly, though. Include something like:
```
    ble_gatts_rw_authorize_reply_params_t reply;
    reply.type = BLE_GATTS_AUTHORIZE_TYPE_READ;
    reply.params.read.gatt_status = BLE_GATT_STATUS_SUCCESS;
    reply.params.read.update = 1;
    reply.params.read.offset = 0;
    reply.params.read.len = 2;
    reply.params.read.p_data = hrmdata;
    sd_ble_gatts_rw_authorize_reply(conn_hdl, &reply);
```
Would that be an acceptable example?

elral:
Hello ptyork,
thank you very much for your example, that made my day!
I was just looking how to react on a read request. After copying your code and adapting to my needs it works. The only thing i figured out is that `setReadAuthorizeCallback` is called twice, first if a read request is issued and second when the actualized data is read. I don't know if this is the "normal" behaviour, but I handle this in my code.
Again thank you and regards
Ralf

hathach:
> Yes, this is complicated, which is why an example would be quite nice per #245. AFAIK, the only way to provide an updated value for a characteristic "on demand" is by responding to a read authorize request. There's no other "I want to read" callback. Which means this is kinda vital to either explain and/or "wrap" in a more elegant way.
> 
> Which actually leads to a question I've got. What is the purpose of this library? Is it just to put an "Adafruit" face on top of the nrf SDK? Or is it to provide a layer of abstraction such that you can later decouple this from Nordic? Obviously if the latter, then you'd definitely want to provide some kind of wrapper around the read and write authorize response process. Even if it is more robust than this particular pull request.
> 
> Anyway, referring back to #245, I'll be glad to submit a pull request that mods custom_hrm to setProperties(CHR_PROPS_READ|CHR_PROPS_NOTIFY) and implement the callback to provide a heart rate when Notify is disabled. Callback would be pretty ugly, though. Include something like:
> 
> ```
>     ble_gatts_rw_authorize_reply_params_t reply;
>     reply.type = BLE_GATTS_AUTHORIZE_TYPE_READ;
>     reply.params.read.gatt_status = BLE_GATT_STATUS_SUCCESS;
>     reply.params.read.update = 1;
>     reply.params.read.offset = 0;
>     reply.params.read.len = 2;
>     reply.params.read.p_data = hrmdata;
>     sd_ble_gatts_rw_authorize_reply(conn_hdl, &reply);
> ```
> 
> Would that be an acceptable example?

@ptyork sorry, somehow this reply of yours fall off my radar.  I am appreciated that you are willing to submit an PR to demonstrate the usage of this authorize callback. However, It would be best to add a new example than modify existing one. Maybe another service that we can tested wtih the nRF Toolbox app

image

quentinmit:
> Hello ptyork,
> thank you very much for your example, that made my day!
> I was just looking how to react on a read request. After copying your code and adapting it to my needs it works. The only thing i figured out is that `setReadAuthorizeCallback` is called twice, first if a read request is issued and second when the actualized data is read. I don't know if this is the "normal" behaviour, but I handle this in my code.
> Again thank you and regards
> Ralf

FYI for anyone else who finds this issue (in lieu of a real example :(), it appears that the `read_authorize_cb` is called for each packet that is sent. I found that I could check for `request->offset == 0` to find the first read, and only update the data in that case. Here's my complete callback function for reference (in this case it's pulling the runtime stats from FreeRTOS):

```c++
void BLERunTimeStats::read_authorize_cb(uint16_t conn_hdl, BLECharacteristic* chr, ble_gatts_evt_read_t * request)
{
  if (Serial) {
    Serial.print("read_authorize_cb(");
    Serial.print(request->offset);
    Serial.println(")");
  }

  ble_gatts_rw_authorize_reply_params_t reply;
  reply.type = BLE_GATTS_AUTHORIZE_TYPE_READ;
  reply.params.read.gatt_status = BLE_GATT_STATUS_SUCCESS;
  reply.params.read.update = 0;
  if (request->offset == 0) {
    char buf[2048];
    vTaskGetRunTimeStats(buf);
    uint16_t len = strlen(buf);
    if (len > BLE_GATTS_VAR_ATTR_LEN_MAX) {
      len = BLE_GATTS_VAR_ATTR_LEN_MAX;
    }
    reply.params.read.update = 1;
    reply.params.read.offset = 0;
    reply.params.read.len = len;
    reply.params.read.p_data = (uint8_t*) buf;
  }
  sd_ble_gatts_rw_authorize_reply(conn_hdl, &reply);
}
```

PR #304, opened 2019-07-10
use code of this repo to run build test instead of release in arduino package

hathach:


ladyada:
good idea!

hathach:
I run into a travis build error due to an modification to the core :D

Issue #307, opened 2019-07-19
[Bug] Serial is not 8 bit clean


pyro9:
Tested on nrf52840 Bluefruit Feather Express rev. D with BSP 0.11.1.

Reading any byte from Serial where the MSB is set is mis-interpreted as no data available.
Adafruit_TinyUSB_Core/Adafruit_USBD_CDC.cpp read returns the value of tud_cdc_read_char() promoted from signed char to signed int.

Since Stream::readBytes calls read, it fails to read the full count when a byte has MSB set.

The following quick and dirty hack correctly reads the full count:

    size_t readBytes(Stream &s, uint8_t *buf, size_t count) {
      uint8_t *pos = buf;

      for(int i = count; i; i--) {
          while(!s.available());
          *pos = s.read();
          pos++;
      }
      return count;
    }

(of course, it doesn't honor timeout since it's a quick and dirty hack)

To get what appears to be the intended behavior throughout the stack, it looks like at least tud_cdc_n_peek and tud_cdc_n_read_char need to return signed int so they can express 0-0xff or -1 correctly.

pyro9:
Patch applied

hathach:
Thanks @pyro9 for finding and fixing this bug

PR #308, opened 2019-07-20
Make Serial::read and dependant functions 8-bit clean.

pyro9:
Fixes issue #307.
Make CDC serial functions that read a single character return an int rather than signed character so -1 for no data available can be distinguished from reading byte 0xFF.

### DIFF
--- a/cores/nRF5/Adafruit_TinyUSB_Core/tinyusb/src/class/cdc/cdc_device.h
+++ b/cores/nRF5/Adafruit_TinyUSB_Core/tinyusb/src/class/cdc/cdc_device.h
@@ -57,10 +57,10 @@ void        tud_cdc_n_get_line_coding (uint8_t itf, cdc_line_coding_t* coding);
 void        tud_cdc_n_set_wanted_char (uint8_t itf, char wanted);
 
 uint32_t    tud_cdc_n_available       (uint8_t itf);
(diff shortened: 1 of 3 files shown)


hathach:
ah thanks, this seems to be a bug with tinyusb stack

cores/nRF5/Adafruit_TinyUSB_Core/tinyusb/src/class/cdc/cdc_device.h
```diff
@@ -57,10 +57,10 @@ void        tud_cdc_n_get_line_coding (uint8_t itf, cdc_line_coding_t* coding);
 void        tud_cdc_n_set_wanted_char (uint8_t itf, char wanted);
 
 uint32_t    tud_cdc_n_available       (uint8_t itf);
-signed char tud_cdc_n_read_char       (uint8_t itf);
+int         tud_cdc_n_read_char       (uint8_t itf);
```
hathach: would you mind changing all `int` to `int32_t`, since this is part of tinyusb stack, I would prefer to stick with known size return value. Since the stack isn't limited to 32-bit mcu.

pyro9: I've added a commit to do int -> int32_t

PR #312, opened 2019-07-27
Added a couple delete operators to make std=gnu++14/17 link

kevinfrei:
Without this change, you'll get link errors for a variety of destructors complaining about "undefined operator delete(void *, unsigned int)" when you try to compile with -std=gnu++14 (or gnu++17) with GCC 7.2.1 (arm-none-eabi-gcc 7-2017q4)


hathach:
thank you very much, I only compiled with gnu++11 so far, isn't aware of this issue :)

Issue #80, opened 2017-10-10
PHY selection support


hathach:
To archive higher throughput for longer range, we need to select approx PHY
related: #82 #83

ref https://www.nordicsemi.com/eng/Products/Bluetooth-5
ref https://www.novelbits.io/bluetooth-5-speed-maximum-throughput/

hathach:
implemented #324

PR #223, opened 2019-01-30
DO NOT MERGE: Supports setting PHY for extended range

mharper:
Adds `setPhy`/`getPhy` for setting PHY. On nRF52840, allows setting PHY
to `BLE_GAP_PHY_CODED` to support extended range.

This work compiles but is not yet field-tested.

Implements #82.

Issue #82, opened 2017-10-10
Bluetooth 5 Long Range support


hathach:
**Longer range**

Bluetooth 5 introduced significant changes at the physical (PHY) layer for Bluetooth radio operation. Four data rates are now available 2Mbps, 1Mbps, 500kbps and 125kbps. The 2Mbps clearly offers higher throughput possibilities. Why offer lower datarates of 500kbps and 125kbps some may ask. The reason for these is that at these lower datarates and with the associated LE Coded processing using Forward Error Correction (FEC) it becomes easier to retirieve data transmissions in surrounding noise. In short, the sensitivity improves, specifically by around 12dB.

In RF terms 6dB improved gain in the link budget is a theoretical doubling of range, therefore 12dB improvement increases range possibilities fourfold. Link budgets can be improved by additional power, but critically the long range improvements in Bluetooth 5 do not rely on improved output power and mean up to four times more range can be achieved without a need to increase peak output power. The data throughput capacity does of course reduce when 500kbps or 125kbps modes are selected, but for many Bluetooth low energy applications which are relatively straightforward sensing and actuating applications, these data rates are perfectly adequate.

lefedor:
https://www.nordicsemi.com/eng/Products/Bluetooth-5
Have interest int this myself.
Want mention that long range is not in the nRF52832 feature list, only for the nRF52840.

hathach:
nrf52840 is coming to Adafruit store soon enough :D

mharper:
I have a relatively immediate need for this feature and have a nRF52840 Feather landing here in a few days.  @hathach if you can give me a vector or two on how to implement this, I can do it.

hathach:
This is new to me as well, you need to dive into Nordic docs and code to get it work. Please let us know what you figure out.

mharper:
Acknowledged. Will let you know what I find out.

mharper:
Based on what I've pieced together so far, the best example of how to do this is in the nRF SDK example found here:
```
examples/ble_central_and_peripheral/experimental/ble_app_att_mtu_throughput
```

Relevant struct: [`ble_gap_phys_t`](https://www.nordicsemi.com/DocLib/Content/SoftDevice_API_Doc/S140/v6-1-0/structble__gap__phys__t)
[...]

PR #324, opened 2019-08-16
Maximize throughput

### LINKED ISSUE #83: High Throughput (2 Mbps PHY) support
ref https://www.nordicsemi.com/eng/Products/Bluetooth-5
ref https://www.novelbits.io/bluetooth-5-speed-maximum-throughput/ 

**Higher data throughput**

The data rate for Bluetooth low energy has now been extended from a singular transmission speed (1Mbs) to four discrete data rate speeds as seen in the list above. The maximum speed is doubled to 2Mbs. Together with advancements introduced in Bluetooth 4.2 which allowed for Data Length Extensions (DLE) overall throughput is x5 higher than the original Bluetooth 4.0 level. Now net data throughput not including overhead such as addressing can be as high as 1.4Mbs. Data packet size, datarate and overall data throughput are illustrated in fig. 1.

### LINKED ISSUE #106: Self initiate MTU exchange
function is `sd_ble_gattc_exchange_mtu_request()` , by BLE specs only can be initiate by client side.

hathach:
- add requestPHY() to switch to 2MB PHY, fix #83  #80
- add requestDataLengthUpdate() to increase data length
- add requestMtuExchange() to actively exchange MTU, fix #106

update throughput example, and added central_throughput example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - BSP_VERSION=`eval ls $BSP_PATH`
   - rm -r $BSP_PATH/*
   - ln -s $TRAVIS_BUILD_DIR $BSP_PATH/$BSP_VERSION
-  - arduino --install-library "Adafruit NeoPixel","Adafruit NeoMatrix","Adafruit GFX Library","Adafruit SSD1306","MIDI Library",
+  - arduino --install-library "Adafruit NeoPixel","Adafruit NeoMatrix","Adafruit GFX Library","Adafruit SSD1306","MIDI Library","Adafruit ILI9341"
 
 before_script:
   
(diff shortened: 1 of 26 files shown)


tilink:
Is Coded PHY (Long Range) supported now because of this? You say in #223 "**implemented** by PR #324" https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/223#issuecomment-522230891 which makes me believe that is the case. However there is another open issue https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/82 that is specifically for Long Range that shows no support so far.

Can I just request PHY Coded the same way you added the 2Mbps support/example?
 `conn->requestPHY(BLE_GAP_PHY_CODED);`

hathach:
@tilink sorry, that was my mistake, just reopen the other issue. Long range with coded phy is not supported yet.

sylvioalves:
Is long range supported at all?

Issue #335, opened 2019-08-29
[Feature] Additional Device Information


elral:
For my project I need to set additional device parameters. To do this I added some methods to BLEDis.cpp and BLEDis.h. It would be great if these additions could be added to the masterbranch.  Unfortenutely I do not know how to submit a pull request, therefore I addded my code here.

Thanks and regards

[BLEDis.zip](https://github.com/adafruit/Adafruit_nRF52_Arduino/files/3554159/BLEDis.zip)

Ralf

hathach:
submit a PR is very simple, it is much easier for you and us than attaching zip file. Just create a folk, modify the code then submit PR. Here is the tutorial from github.
https://help.github.com/en/articles/creating-a-pull-request

elral:
Thanks, I forked it and changed the files, but "unfortenutely" more than for this Pull Request. It seems that no files can be choosen for a Pull Request, so I have to start again. Will start next week.

hathach:
You dont need to choose file, only need to modify, add file to github then commit and push. Then create the pr, github will do the rest for you

PR #336, opened 2019-09-02
Additional Device Informations for ble

elral:
Added some methods for setting more device informations.
See also Issue #335

### DIFF
--- a/libraries/Bluefruit52Lib/src/services/BLEDis.cpp
+++ b/libraries/Bluefruit52Lib/src/services/BLEDis.cpp
@@ -51,26 +51,111 @@ BLEDis::BLEDis(void)
   _hardware_rev = NULL;
   _software_rev = ARDUINO_BSP_VERSION;
   _manufacturer = "Adafruit Industries";
+  _system_id    = NULL;
+  _reg_cert_list = NULL;
+  _pnp_id       = NULL;
+}
+
+void BLEDis::setFirmwareRev(const char* firmware_rev)
+{
+  _firmware_rev = firmware_rev;
+  _firmware_rev_length = strlen(firmware_rev);
+}
+void BLEDis::setFirmwareRev(const char* firmware_rev, uint8_t length)
+{
+  _firmware_rev = firmware_rev;
+  _firmware_rev_length = length;
+}
+
+void BLEDis::setSerialNum(const char* serial_num)
+{
+  _serial = serial_num;
+  _serial_length = strlen(serial_num);
+}
+void BLEDis::setSerialNum(const char* serial_num, uint8_t length)
+{
+  _serial = serial_num;
+  _serial_length = length;
+}
+
+void BLEDis::setSystemID(const char* system_id)
+{
+  _system_id = system_id;
+  _system_id_length = strlen(system_id);
+}
+void BLEDis::setSystemID(const char* system_id, uint8_t length)
+{
+  _system_id = system_id;
+  _system_id_length = length;
+}
+
+void BLEDis::setRegCertList(const char* reg_cert_list)
+{
+  _reg_cert_list = reg_cert_list;
+  _reg_cert_list_length = strlen(reg_cert_list);
+}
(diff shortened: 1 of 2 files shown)


libraries/Bluefruit52Lib/src/services/BLEDis.h
```diff
@@ -46,24 +46,56 @@ class BLEDis : public BLEService
...
+        uint8_t _pnp_id_length;
+      };
+      const uint8_t _strarr_length[9];
     };
 
   public:
     BLEDis(void);
 
     void setModel(const char* model);
+    void setModel(const char* model,uint8_t length);
```
hathach: for using this API with length, I would suggest to 
1. change the type of 1st parameter to `const uint8_t*` e.g `setModel(const uint8_t model, uint8_t length)`
2. also please implement the `const char*` as inline function as well.

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
```diff
@@ -51,51 +51,164 @@ BLEDis::BLEDis(void)
...
     {
-      BLECharacteristic chars(UUID16_CHR_MODEL_NUMBER_STRING+i);
+      BLECharacteristic chars(UUID16_CHR_SYSTEM_ID+i);
       chars.setTempMemory();
-
       chars.setProperties(CHR_PROPS_READ);
-      chars.setFixedLen(strlen(_strarr[i]));
-
+      if (_strarr_length[i]) {
+        chars.setFixedLen(_strarr_length[i]);
```
hathach: 3. should only use the _strarr_legnth[i], there is no point to use the strlen() function anymore.

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
```diff
@@ -51,51 +51,164 @@ BLEDis::BLEDis(void)
...
+      } else {
+        chars.setFixedLen(strlen(_strarr[i]));
+      }
       VERIFY_STATUS( chars.begin() );
-      chars.write(_strarr[i]);
+      if (_strarr_length[i]) {
```
hathach: same as above

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
```diff
@@ -51,51 +51,164 @@ BLEDis::BLEDis(void)
...
+        chars.write(_strarr[i], _strarr_length[i]);
+      } else {
+        chars.write(_strarr[i]);
+      }
+      
     }
   }
 
+  if ( _strarr[arrcount(_strarr)-1] != NULL )
```
hathach: 4. I am not familliar with PNP_ID, but why this characteristic isn't included in the above loop

libraries/Bluefruit52Lib/src/services/BLEDis.h
elral: 1. Yes, you are right. Then inside function it must be "_model = (const char*)model;"
2. Sorry, I don't know what to change, or do you mean "_model = (const char*)model;"

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
elral: Yes, you are rigth.

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
elral: Yes, you are rigth.

libraries/Bluefruit52Lib/src/services/BLEDis.cpp
elral: Last characteristic from the loop is 0x2A2A (IEEE_REGULATORY_CERTIFICATION), PNP_ID is 0x2A50. So there is a big gap between last entry from loop and PNP_ID. I had no better idea than to do it seperatly.

hathach: ah i see the reason now.

Issue #330, opened 2019-08-21
[Feature] Updating advertising data on the fly


martin2250:
It would be useful if there was an option to change advertising data on the fly (without stopping and re-starting advertising).
`sd_ble_gap_adv_set_configure` can be called even when advertising is active, so my suggestion is to add a new method `updateData()` to `BLEAdvertising`, that more or less contains this code: https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/d3f3b34319fefddf8ac67bd4a0587e189ab4cbb8/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp#L341-L363

Another solution would be to have `BLEAdvertising::_start` check if advertising is already active

elral:
I have done this in a separate method, but following has to considered:
` * @param[in]     p_adv_params                         Advertising parameters. When this function is used to update advertising data while advertising, this parameter must be NULL. See @ref ble_gap_adv_params_t.`
That means it must be `VERIFY_STATUS( sd_ble_gap_adv_set_configure(&_hdl, &gap_adv, NULL), false );`

PR #337, opened 2019-09-02
Update Advertising data on the fly

elral:
Before calling this function advertising data has to be set using "Bluefruit.Advertising.setData(data,len" and/or "Bluefruit.ScanResponse.setData(data,len)".
See also Issue #330

### DIFF
--- a/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
+++ b/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
@@ -385,6 +385,21 @@ bool BLEAdvertising::start(uint16_t timeout)
   return true;
 }
 
+// first set new advertising data with "Bluefruit.Advertising.setData(data,len) or "Bluefruit.ScanResponse.setData(data,len)"
+bool BLEAdvertising::updateData()
+{
+  // gap_adv long-live is required by SD v6
+  static ble_gap_adv_data_t gap_adv =
+  {
+      .adv_data      = { .p_data = _data, .len = _count },
+      .scan_rsp_data = { .p_data = Bluefruit.ScanResponse.getData(), .len = Bluefruit.ScanResponse.count() }
+  };
+  VERIFY_STATUS( sd_ble_gap_adv_set_configure(&_hdl, &gap_adv, NULL), false );
+  VERIFY_STATUS( sd_ble_gap_adv_start(_hdl, CONN_CFG_PERIPHERAL), false );
+
+  return true;
+}
+
 bool BLEAdvertising::stop(void)
 {
   VERIFY_STATUS( sd_ble_gap_adv_stop(_hdl), false);
@@ -465,4 +480,3 @@ void BLEAdvertising::_eventHandler(ble_evt_t* evt)
     default: break;
   }
 }
-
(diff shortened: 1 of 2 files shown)


zwells:
Will this work with  Bluefruit.ScanResponse.addManufacturerData(data, len); ?

zwells:
Nevermind, this works without the pull request:
  Bluefruit.ScanResponse.clearData();
  Bluefruit.ScanResponse.addManufacturerData(manufacturerData,sizeof(manufacturerData));

libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
```diff
@@ -385,6 +385,21 @@ bool BLEAdvertising::start(uint16_t timeout)
   return true;
 }
 
+// first set new advertising data with "Bluefruit.Advertising.setData(data,len) or "Bluefruit.ScanResponse.setData(data,len)"
+bool BLEAdvertising::updateData()
+{
+  // gap_adv long-live is required by SD v6
+  static ble_gap_adv_data_t gap_adv =
```
hathach: this will create an extra memory for gap_adv, we should move it to class's member. Though I could do this myself

libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
```diff
@@ -385,6 +385,21 @@ bool BLEAdvertising::start(uint16_t timeout)
...
+  {
+      .adv_data      = { .p_data = _data, .len = _count },
+      .scan_rsp_data = { .p_data = Bluefruit.ScanResponse.getData(), .len = Bluefruit.ScanResponse.count() }
+  };
+  VERIFY_STATUS( sd_ble_gap_adv_set_configure(&_hdl, &gap_adv, NULL), false );
+  VERIFY_STATUS( sd_ble_gap_adv_start(_hdl, CONN_CFG_PERIPHERAL), false );
```
hathach: can you verify that we can just call the `set_configure()` while advertising, and the data will change without us to call `adv_start()`. Otherwise I think we loose the meaning of on-the-fly.

elral: Sorry for the late answer, I was on vacation with limited access.
I use the proposed changes for several weeks, so after starting advertising just calling sd_ble_gap_adv_set_configure(...) will update the data.

elral: I must correct my answer. After rechecking my code I was still using adv_start() inside updateData(). Without adv_start() it does not work.

As mentioned by zwells it works for updating the data by using:
Bluefruit.ScanResponse.clearData();
Bluefruit.ScanResponse.addManufacturerData(manufacturerData,sizeof(manufacturerData));

But the advertising type and length must not be changed, otherwise you will get an invalid EIR (as per nrfConnect). Unfortunately it does also not work using the proposed changes.
So for switching the scanRespanse data between e.g. manufactureName and manufactureData will not work and up to now I didn't find a solution.

Sorry for the confusion and regards

Ralf

hathach: Sorry, I am late again, thanks @elral for your effort and PR. We close since the PR doesn't resolve the on-the-fly advertising. It is better to open a new PR later on

PR #356, opened 2019-10-14
Rename homekit_lightbulb.ino.cpp to homekit_lightbulb.ino

tamberg:


ladyada:
thank you!

hathach:
@ladyada This HomeKit example is no longer  compile with current BSP. This will break travis build, we need to add skip file to skip travis build. 

Previously this sketch only passed the pairing (adding new device) process. I didnt have time to get it working with transaction message yet.

ladyada:
ooh ok i think maybe better to put test skip files in and put big warning at the top, then?

PR #358, opened 2019-10-17
clean up

hathach:
- update nrfx to 1.7.2
- clean up defines
- clean up lfs flash


ladyada:
i tested this with CPB and feather nrf52832, various examples

PR #360, opened 2019-10-17
use SPIM3 for nrf52840 with maximum speed at 32 Mhz

### LINKED ISSUE #351: [Feature] allow SPI creation with high speed SPIM3
perhaps in the variant you can specify you want SPI/SPI1/etc to use the high speed SPIM3 periph. then we can have 32mhz instead of 8mhz for TFT for easy 4x speedup :)

### LINKED ISSUE #354: replace SPI by SPIM

hathach:
- close #354 
- close #351
due to 832 errata, SPIM is only used for 840, 832 still use the legacy SPI with non-dma https://infocenter.nordicsemi.com/topic/errata_nRF52832_Rev2/ERR/nRF52832/Rev2/latest/anomaly_832_58.html


ladyada:
ok verified with TFT gizmo + CPB, its slower (bigger delays between bytes) but they are 16mhz

ladyada:
also tested feather nrf52840 w/wing to test LCD out and touchscreen in!

PR #372, opened 2019-11-04
[SUPERCEDED by #397] Guard against race conditions in flash FS cache

### LINKED ISSUE #222: Connecting/Reconnecting a nrf52 peripheral to a central can corrupt the LittleFS FS where pairing information is stored.
- Please provide details that help us to reproduce your issue
That's a hard one... generally disconnecting and reconnecting can cause this issue but doesn't happen all the time.  It's not a critical problem when a central (nrf52832) reconnects to a peripheral (nrf52832) but is an issue when a computer reconnects to a "faulty" peripheral.  It appears that a nrf52 central will re-negociate the pairing information.  However, a computer will retain the pairing information and keep retrying (as seen by the toggling back and forth between paired and connected on windows 10).

I suspect the issue occurs on the original pairing and not the reconnection.

  - Your Board and BSP version
       BSP Library : 0.9.3
       Bootloader  : s132 6.1.1 r1
  - Your OS: 
       Windows 10
  - Steps to reproduce, or better your (simplified) sketch if possible
Can't yet identify the steps to reproduce or the root cause. (will keep working on that)

  - Serial debug log: all at Debug Level 2:
Problem 1 (on reconnection - but not the actual root cause):

```
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::rmdir_r(const char*): 373: verify failed, error = LFS_ERR_CORRUPT

```

Problem 2 (after restarting):

```
Peripheral Paired Devices: 
lfs error:494: Corrupted dir pair at 124 123
BluefruitFS::File LittleFS::_open_dir(const char*): 281: verify failed, error = LFS_ERR_CORRUPT

```

When windows tries to reconnect, all we see is this (and it keeps repeating):

```
[BLE   ] BLE_GAP_EVT_CONNECTED : Conn Handle = 0
[GAP   ] Conn Interval= 11.250000
[BLE   ] BLE_GAP_EVT_SEC_INFO_REQUEST : Conn Handle = 0
lfs error:494: Corrupted dir pair at 124 123
virtual BluefruitFS::File LittleFS::open(const char*, uint8_t): 313: verify failed, error = LFS_ERR_CORRUPT
[BLE   ] BLE_GAP_EVT_DISCONNECTED : Conn Handle = 0
[GAP   ] Disconnect Reason 0x13
```

Even after loading the "Clear Bonds" program. I was left with the same problem:

```
Bluefruit52 Clear Bonds Example
-------------------------------

[CFG   ] SoftDevice's RAM requires: 0x20002530
lfs error:494: Corrupted dir pair at 3 2
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::mkdir(const char*): 339: verify failed, error = LFS_ERR_CORRUPT
lfs error:494: Corrupted dir pair at 3 2
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::mkdir(const char*): 339: verify failed, error = LFS_ERR_CORRUPT

----- Before -----

lfs error:494: Corrupted dir pair at 3 2
virtual BluefruitFS::File LittleFS::open(const char*, uint8_t): 313: verify failed, error = LFS_ERR_CORRUPT

lfs error:494: Corrupted dir pair at 3 2
virtual BluefruitFS::File LittleFS::open(const char*, uint8_t): 313: verify failed, error = LFS_ERR_CORRUPT

lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::rmdir_r(const char*): 373: verify failed, error = LFS_ERR_CORRUPT
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::mkdir(const char*): 339: verify failed, error = LFS_ERR_CORRUPT
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::rmdir_r(const char*): 373: verify failed, error = LFS_ERR_CORRUPT
lfs error:494: Corrupted dir pair at 3 2
virtual bool LittleFS::mkdir(const char*): 339: verify failed, error = LFS_ERR_CORRUPT

----- After  -----

lfs error:494: Corrupted dir pair at 3 2
virtual BluefruitFS::File LittleFS::open(const char*, uint8_t): 313: verify failed, error = LFS_ERR_CORRUPT

lfs error:494: Corrupted dir pair at 3 2
virtual BluefruitFS::File LittleFS::open(const char*, uint8_t): 313: verify failed, error = LFS_ERR_CORRUPT
```

At that point, the only way I was able to recover FS functionality was to re-flash the bootloader and re-flash clearbonds which gave me: 

```
[CFG   ] SoftDevice's RAM requires: 0x20002530
lfs error:494: Corrupted dir pair at 0 1
lfs error:2222: Invalid superblock at 0 1
[IFLASH] Format internal file system
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS

----- Before -----



[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS

----- After  -----

```

After reflashing over the reformatted file system, it works fine, until the offending situation occurs and the problem happens again.

When the bonding information is wiped by reflashing the bootloader, I can unpair and reconnect in windows 10 and I get these messages when pairing:

```

[GAP   ] Conn Interval= 11.250000
[BLE   ] BLE_GATTC_EVT_CHAR_VAL_BY_UUID_READ_RSP : Conn Handle = 1
[BLE   ] BLE_GAP_EVT_SEC_PARAMS_REQUEST : Conn Handle = 1
[PRPH  ] Connected to 1 TOASTER
[BLE   ] BLE_GATTS_EVT_EXCHANGE_MTU_REQUEST : Conn Handle = 1
[GAP   ] ATT MTU is changed to 23
[BLE   ] BLE_GAP_EVT_CONN_SEC_UPDATE : Conn Handle = 1
[BOND  ] CCCD setting not found
[BLE   ] BLE_GAP_EVT_AUTH_STATUS : Conn Handle = 1
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BLE   ] BLE_GATTC_EVT_CHAR_VAL_BY_UUID_READ_RSP : Conn Handle = 1
[BOND  ] Saved keys for "TOASTER" to file /adafruit/bond_prph/59d1 ( 90 bytes )
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BLE   ] BLE_GATTS_EVT_WRITE : Conn Handle = 1
[BLE   ] BLE_GAP_EVT_CONN_PARAM_UPDATE : Conn Handle = 1
[GAP   ] Conn Interval= 11.250000
[BLE   ] BLE_GATTS_EVT_WRITE : Conn Handle = 1
[GATTS ] attr's cccd
000:  01 00                                           | ..

[BOND  ] Saved CCCD setting to file /adafruit/bond_prph/59d1 ( offset = 90, len = 44 bytes )
[BLE   ] BLE_GATTS_EVT_WRITE : Conn Handle = 1
[GATTS ] attr's cccd
000:  01 00                                           | ..

[BLE   ] BLE_GAP_EVT_CONN_PARAM_UPDATE : Conn Handle = 1
[GAP   ] Conn Interval= 56.250000
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BLE   ] BLE_GATTS_EVT_WRITE : Conn Handle = 1
[GATTS ] attr's cccd
000:  01 00                                           | ..

[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BOND  ] Saved CCCD setting to file /adafruit/bond_prph/59d1 ( offset = 90, len = 44 bytes )
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BLE   ] BLE_GAP_EVT_CONN_PARAM_UPDATE : Conn Handle = 1
[GAP   ] Conn Interval= 11.250000
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BOND  ] Saved CCCD setting to file /adafruit/bond_prph/59d1 ( offset = 90, len = 44 bytes )
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
[BLE   ] BLE_GATTS_EVT_WRITE : Conn Handle = 1
[GATTS ] attr's value, uuid = 0x2A4D
000:  01                                              | .

[HID   ] Keyboard LED : 0x01
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS
```

### LINKED ISSUE #227: InternalFS Corrupt Blocks
Board: nRF52840 Express
MacOS
Bootloader 0.2.6
Arduino IDE 1.8.8

I'm trying to write a string to the InternalFS, and I get the following error sporadically.
```
12:59:53.579 -> lfs error:494: Corrupted dir pair at 0 1
12:59:53.579 -> lfs error:2222: Invalid superblock at 0 1
12:59:53.579 -> [IFLASH] Format internal file system
```
The only way to fix it is to InternalFS.format(true).  I can't quite pinpoint what is causing it, so I don't think posting code will help.  Having said that, I did some research and discovered the following which may/may not help:

1) InternalFS File.write(buffer, size) does not overwrite the file. It appends to it.  I incremented a counter every 1s and read/write it to the same file and got the following output:
```
1
12
123
```
Instead, I would expect the following:
```
1
2
3
```
Which takes me to 2). 

2)  In order to overwrite the file contents, I remove the file and create a new one. Removing a file takes a VERY long time (a few seconds).  I found the following article from Nordic that describes how internal flash is slow to remove:

https://devzone.nordicsemi.com/f/nordic-q-a/20197/nrf52-flash-memory-vs-eeprom

It could be that since I'm running BLE and removing files, RTOS is switching tasks which is causing issues?  Anyways, right now I've created an RTOS queue to serialize my writes, and/or maybe I need to add a mutex.  I think once ExternalFS is up and running these kind of problems go away.. but I thought I'd post my findings here.

### LINKED ISSUE #325: nRF52832 frequently connect/disconnect occur assertion "head >= 2 && head <= lfs->cfg->block_count"
**Hi~
I'm using a nRF52832 feather to build keyboard. I found a strange bug when iPhone disconnecting and reconnecting keyboard frequently can cause error:**

10:36:26.458 -> assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "C:\Users\cc\AppData\Local\Arduino15\packages\adafruit\hardware\nrf52\0.11.1\libraries\Adafruit_LittleFS\src\littlefs\lfs.c", line 1144, function: lfs_ctz_find
10:36:33.668 -> assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "C:\Users\cc\AppData\Local\Arduino15\packages\adafruit\hardware\nrf52\0.11.1\libraries\Adafruit_LittleFS\src\littlefs\lfs.c", line 1144, function: lfs_ctz_find

**In my project, iPhone is peripheral and keyboard is cent, iPhone connect keyboard when it started. The connect/disconnect callback code like this:**

```
void prphConnectCallback(uint16_t conn_handle) {
  // Get the reference to current connection
  BLEConnection* connection = Bluefruit.Connection(conn_handle);
  perph_conn_handle = conn_handle;

  // debug information
  char peer_name[32] = { 0 };
  connection->getPeerName(peer_name, sizeof(peer_name));

  LOG("[Prph] Connected handle: %d, name: %s\r\n", conn_handle, peer_name);
}

void prphDisconnectCallback(uint16_t conn_handle, uint8_t reason) {
  (void) conn_handle;
  (void) reason;

  Bluefruit.Advertising.stop();
  delay(10);

  //startAdv();
  Bluefruit.Advertising.start(0);
  // debug information
  LOG("[Prph] Disconnected\r\n");
}
```

**The complete serial logs is(Level 0):**
02:31:05.162 -> start...
02:31:05.435 -> in low power
02:31:05.468 -> LN-REDOX bluetooth keyboard RHS started
02:31:05.876 -> [SCAN] scanned, start connect...
02:31:06.186 -> [Cent] Connected handle: 0, name: LHS
02:31:10.046 -> [Prph] Connected handle: 1, name: iPhone
02:31:22.517 -> [Prph] Disconnected
02:31:26.375 -> [Prph] Connected handle: 1, name: iPhone (2
02:31:41.828 -> [Prph] Disconnected
02:31:47.139 -> [Prph] Connected handle: 1, name: iPhone (2
02:31:59.740 -> [Prph] Disconnected
02:32:03.654 -> [Prph] Connected handle: 1, name: iPhone (2
02:32:31.155 -> [Prph] Disconnected
02:32:35.173 -> [Prph] Connected handle: 1, name: iPhone (2
02:32:52.566 -> [Prph] Disconnected
02:32:56.388 -> [Prph] Connected handle: 1, name: iPhone (2
02:33:11.480 -> [Prph] Disconnected
02:33:14.658 -> [Prph] Connected handle: 1, name: iPhone (2
02:33:27.269 -> [Prph] Disconnected
02:33:31.146 -> [Prph] Connected handle: 1, name: iPhone (2
02:33:56.366 -> [Prph] Disconnected
02:34:03.249 -> [Prph] Connected handle: 1, name: iPhone (2
02:34:04.756 -> [Prph] Disconnected
02:34:16.321 -> [Prph] Connected handle: 1, name: iPhone (2
02:37:26.456 -> [Prph] Disconnected
02:42:45.820 -> [Prph] Connected handle: 1, name: iPhone (2
02:53:56.825 -> [Prph] Disconnected
08:45:48.984 -> [Prph] Connected handle: 1, name: 
09:03:59.574 -> [Prph] Disconnected
09:33:22.403 -> [Prph] Connected handle: 1, name: iPhone (2
09:34:09.581 -> [Prph] Disconnected
10:03:07.080 -> [Prph] Connected handle: 1, name: iPhone (2
10:03:28.158 -> [Prph] Disconnected
10:23:34.210 -> [Prph] Connected handle: 1, name: 
10:23:34.659 -> [Prph] Disconnected
10:23:42.526 -> [Prph] Connected handle: 1, name: 
10:23:42.526 -> [Prph] Disconnected
10:23:43.323 -> [Prph] Connected handle: 1, name: 
10:23:44.549 -> [Prph] Disconnected
10:23:49.709 -> [Prph] Connected handle: 1, name: 
10:23:51.040 -> [Prph] Disconnected
10:23:53.630 -> [Prph] Connected handle: 1, name: 
10:23:54.343 -> [Prph] Disconnected
10:27:51.664 -> [Prph] Connected handle: 1, name: 
10:27:52.276 -> [Prph] Disconnected
10:27:55.238 -> [Prph] Connected handle: 1, name: 
10:27:55.272 -> [Prph] Disconnected
10:27:56.631 -> [Prph] Connected handle: 1, name: iPhone (2
10:28:13.872 -> [Prph] Disconnected
10:28:26.004 -> [Prph] Connected handle: 1, name: iPhone (2
10:28:27.223 -> [Prph] Disconnected
10:28:32.137 -> [Prph] Connected handle: 1, name: iPhone (2
10:28:33.885 -> [Prph] Disconnected
10:36:19.230 -> [Prph] Connected handle: 1, name: iPhone (2
10:36:26.458 -> assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "C:\Users\cc\AppData\Local\Arduino15\packages\adafruit\hardware\nrf52\0.11.1\libraries\Adafruit_LittleFS\src\littlefs\lfs.c", line 1144, function: lfs_ctz_find
10:36:33.668 -> assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "C:\Users\cc\AppData\Local\Arduino15\packages\adafruit\hardware\nrf52\0.11.1\libraries\Adafruit_LittleFS\src\littlefs\lfs.c", line 1144, function: lfs_ctz_find

**I want to know when this error will occur and how to solve it.**

**bootloader: feather_nrf52832_bootloader-0.2.11_s132_6.1.1
bsp: 0.11.1**

@hathach @jeremypoulter 
Plz~

### LINKED ISSUE #350: [Bug] InternalFS analysis: LFS requirements, coherency guarantees
<details>
<summary>original analysis -- useful but may not be cause</summary>

<P>

**Describe the bug**
[ARMmbed/littlefs](https://github.com/ARMmbed/littlefs) allows the `prog` and `erase` functions to return LFS_ERR_CORRUPT.  However, it does not allow the underlying implementation to ignore errors.  Unfortunately, it appears that there are multiple paths where Adafruit's InternalFS can hit an error, but not report that error to LFS.  This bug opened to have clear bug title corresponding to underlying issue.

This thread is based on manual code review and analysis of **BSP v0.14.0**.
Corruptions have been reported on both **nRF52832** and **nRF52840**.
See linked issue #325 for an example serial debug log that shows the relevant code paths were hit just prior to a corruption being detected.  Relevant part:

**11:07:35.692** -> [SOC ] **NRF_EVT_FLASH_OPERATION_ERROR**
11:07:36.308 -> [BLE ] BLE_GAP_EVT_DISCONNECTED : Conn Handle = 1
11:07:36.308 -> [GAP ] Disconnect Reason 0x08
11:07:36.308 -> [SOC ] NRF_EVT_FLASH_OPERATION_SUCCESS
11:07:36.475 -> [SOC ] NRF_EVT_FLASH_OPERATION_SUCCESS
**11:07:36.509** -> assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "C:\Users\cc\AppData\Local\Arduino15\packages\adafruit\hardware\nrf52\0.11.1\libraries\Adafruit_LittleFS\src\littlefs\lfs.c", line 1144, function: lfs_ctz_find

**To Reproduce**
Manually review the code.  Note the multiple code paths where an error is ignored.  Details follow for easy reference (also written at #325).

**Expected behavior**
When an error is reported when programming or erasing the flash, InternalFS must either retry the operation (e.g., up to 5x), or report that the write/erase/flush operation has failed.  This will allow LittleFS to avoid using that page of cache, and maintain data consistency..

**Additional context**
There is a history of hard-to-reproduce file system corruptions (e.g., #325, #227, #226, #222).  Not all have indications of this bug, but then the current BSP has commented-out the trace when errors are reported.

</details>

See [below](https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/350#issuecomment-540158177) for update

henrygab:
Serialize all the flash_cache read/write/flush functions.

This is necessary because these functions may be called by multiple tasks simultaneously, and because these functions use and modify shared state.

@hathach, @ladyada -- This change compiles, but I **_recommend your validation_**, as I have been unable to do so myself.  That said, this has a very strong likelihood of resolving the file system corruption issues.  Testing should include repeated BLE pairing and main sketch file system use.

After additional testing, this PR should fix the following issues:
Fix #350
Fix #325
Fix #227
Fix #222

### DIFF
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -46,13 +46,20 @@ void setup()
   InternalFS.format();
   Serial.println("Done"); Serial.flush();
 
+  // Create a folder for each thread
+  InternalFS.mkdir("high");
+  InternalFS.mkdir("n1");
+  InternalFS.mkdir("n2");
+  InternalFS.mkdir("n3");
+  InternalFS.mkdir("loop");
+
   // Create thread with different priority
   // Although all the thread share loop() code, they are separated threads
   // and running with different priorities
 
   // Note: default loop() is running at LOW
   //Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGHEST, "highest");
-  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGH, "high");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGH  , "high");
   Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n1");
   Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n2");
   Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n3");
@@ -60,8 +67,8 @@ void setup()
 
 void write_files(const char * name)
 {
-  char fname[20] = { 0 };
-  sprintf(fname, "%s.txt", name);
+  char fname[30] = { 0 };
+  sprintf(fname, "%s/%s.txt", name, name); // each task has its own folder and file
 
   File file(InternalFS);
 
@@ -77,26 +84,34 @@ void write_files(const char * name)
 
 void list_files(void)
 {
-  File dir("/", FILE_O_READ, InternalFS);
+  File root("/", FILE_O_READ, InternalFS);
+  File subdir(InternalFS);
   File file(InternalFS);
 
-  while( (file = dir.openNextFile(FILE_O_READ)) )
+  while( (subdir = root.openNextFile(FILE_O_READ)) )
   {
-    if ( file.isDirectory() ) continue;
-
-    Serial.printf("--- %s ---\n", file.name());
-
-    while ( file.available() )
+    if ( subdir.isDirectory() )
     {
-      char buffer[64] = { 0 };
-      file.read(buffer, sizeof(buffer)-1);
+      char fname[30];
+      sprintf(fname, "%s/%s.txt", subdir.name(), subdir.name());
 
-      Serial.print(buffer);
-      delay(100);
+      if ( file.open(fname, FILE_O_READ) )
+      {
+        Serial.printf("--- %s ---\n", fname);
+
+        while ( file.available() )
+        {
+          char buffer[64] = { 0 };
+          file.read(buffer, sizeof(buffer)-1);
+
+          Serial.print(buffer);
+          delay(100);
+        }
+        file.close();
+
+        Serial.println("---------------\n");
+      }
     }
-    file.close();
-
-    Serial.println("---------------\n");
   }
 }
 
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -105,13 +105,15 @@ void list_files(void)
           file.read(buffer, sizeof(buffer)-1);
 
           Serial.print(buffer);
-          delay(100);
+          delay(10);
         }
         file.close();
 
         Serial.println("---------------\n");
       }
     }
+
+    subdir.close();
   }
 }
 
@@ -129,7 +131,7 @@ void loop()
       list_files();
     }
 
-    delay(100);
+    delay(1000);
     vTaskSuspend(NULL); // suspend task
     return;
   }
@@ -142,5 +144,5 @@ void loop()
   write_files(name);
 
   // lower delay increase chance for high prio task preempt others.
-  delay(500);
+  delay(100);
 }
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -25,7 +25,7 @@ using namespace Adafruit_LittleFS_Namespace;
  */
 
 // timeout in seconds
-#define TIME_OUT      600
+#define TIME_OUT      20
 
 uint32_t writeCount = 0;
 
@@ -51,11 +51,11 @@ void setup()
   // and running with different priorities
 
   // Note: default loop() is running at LOW
-  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
-  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
-  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
-  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGH, "high");
   //Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGHEST, "highest");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGH, "high");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n1");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n2");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "n3");
 }
 
 void write_files(const char * name)
@@ -67,7 +67,7 @@ void write_files(const char * name)
 
   if ( file.open(fname, FILE_O_WRITE) )
   {
-    file.printf("%s %d\n", name, writeCount++);
+    file.printf("%d\n", writeCount++);
     file.close();
   }else
   {
@@ -114,6 +114,7 @@ void loop()
       list_files();
     }
 
+    delay(100);
     vTaskSuspend(NULL); // suspend task
     return;
   }
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -28,7 +28,7 @@ using namespace Adafruit_LittleFS_Namespace;
  */
 
 // timeout in seconds
-#define TIME_OUT      30
+#define TIME_OUT      60
 
 // Limit number of writes since our internal flash is limited (28 KB in total)
 #define WRITE_COUNT   2000
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -28,7 +28,7 @@ using namespace Adafruit_LittleFS_Namespace;
  */
 
 // timeout in seconds
-#define TIME_OUT      60
+#define TIME_OUT      30
 
 // Limit number of writes since our internal flash is limited (28 KB in total)
 #define WRITE_COUNT   2000
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -155,7 +155,7 @@ void loop()
       list_files();
     }
 
-    delay(1000);uint32_t duration_ms = 0;
+    delay(1000);
     vTaskSuspend(NULL); // suspend task
     return;
   }
@@ -165,7 +165,6 @@ void loop()
   Serial.flush();
 
   // Write files
-  uint32_t duration_ms = 0;
   write_files(name);
 
   // lower delay increase chance for high prio task preempt others.
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -25,9 +25,9 @@ using namespace Adafruit_LittleFS_Namespace;
  */
 
 // timeout in seconds
-#define TIME_OUT      5
+#define TIME_OUT      600
 
-bool summarized = false;
+uint32_t writeCount = 0;
 
 // the setup function runs once when you press reset or power the board
 void setup() 
@@ -65,7 +65,7 @@ void write_files(const char * name)
 
   if ( file.open(fname, FILE_O_WRITE) )
   {
-    file.println(name);
+    file.printf("%d\n", writeCount++);
     file.close();
   }else
   {
@@ -91,11 +91,11 @@ void list_files(void)
       readlen = file.read(buffer, sizeof(buffer));
 
       Serial.print(buffer);
+      delay(100);
     }
     file.close();
 
     Serial.println("---------------\n");
-    Serial.flush();
   }
 }
 
@@ -107,9 +107,9 @@ void loop()
   if ( millis() > TIME_OUT*1000 )
   {
     // low priority task print summary
-    if ( !summarized && (TASK_PRIO_LOW == uxTaskPriorityGet(th)))
+    if (TASK_PRIO_LOW == uxTaskPriorityGet(th))
     {
-      summarized = true;
+      Serial.printf("Total write count = %d\n", writeCount);
       list_files();
     }
 
@@ -125,5 +125,5 @@ void loop()
   write_files(name);
 
   // lower delay increase chance for high prio task preempt others.
-  delay(100);
+  delay(500);
 }
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -28,7 +28,7 @@ using namespace Adafruit_LittleFS_Namespace;
  */
 
 // timeout in seconds
-#define TIME_OUT      600
+#define TIME_OUT      60
 
 // Limit number of writes since our internal flash is limited (28 KB in total)
 #define WRITE_COUNT   2000
@@ -151,7 +151,7 @@ void loop()
     // low priority task print summary
     if (TASK_PRIO_LOW == uxTaskPriorityGet(th))
     {
-      Serial.printf("Total write count = %d in %f seconds\n", writeCount, millis()/1000);
+      Serial.printf("Total write count = %d in %.2f seconds\n", writeCount, millis()/1000.0);
       list_files();
     }
 
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -1,0 +1,129 @@
+/*********************************************************************
+ This is an example for our nRF52 based Bluefruit LE modules
+
+ Pick one up today in the adafruit shop!
+
+ Adafruit invests time and resources providing this open source code,
+ please support Adafruit and open-source hardware by purchasing
+ products from Adafruit!
+
+ MIT license, check LICENSE for more information
+ All text above, and the splash screen below must be included in
+ any redistribution
+*********************************************************************/
+
+#include <Adafruit_LittleFS.h>
+#include <InternalFileSystem.h>
+
+using namespace Adafruit_LittleFS_Namespace;
+
+/* This example perform a stress test on Little FileSystem on internal flash.
+ * There are 4 different thread sharing same loop() code with different priority
+ *    loop (low), normal, high, highest
+ * Each will open and write a file of its name (+ .txt). Task takes turn writing until timeout
+ * to print out the summary
+ */
+
+// timeout in seconds
+#define TIME_OUT      5
+
+bool summarized = false;
+
+// the setup function runs once when you press reset or power the board
+void setup() 
+{
+  Serial.begin(115200);
+  while ( !Serial ) yield();   // for nrf52840 with native usb
+
+  Serial.println("Internal Stress Test Example");
+  yield();
+
+  // Initialize Internal File System
+  InternalFS.begin();
+
+  // Format
+  Serial.print("Formatting ... "); Serial.flush();
+  InternalFS.format();
+  Serial.println("Done"); Serial.flush();
+
+  // Create thread with different priority
+  // Although all the thread share loop() code, they are separated threads
+  // and running with different priorities
+
+  // Note: default loop() is running at LOW
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGH, "high");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_HIGHEST, "highest");
+}
+
+void write_files(const char * name)
+{
+  char fname[20] = { 0 };
+  sprintf(fname, "%s.txt", name);
+
+  File file(InternalFS);
+
+  if ( file.open(fname, FILE_O_WRITE) )
+  {
+    file.println(name);
+    file.close();
+  }else
+  {
+    Serial.printf("Failed to open %s\n", fname);
+  }
+}
+
+void list_files(void)
+{
+  File dir("/", FILE_O_READ, InternalFS);
+  File file(InternalFS);
+
+  while( (file = dir.openNextFile(FILE_O_READ)) )
+  {
+    if ( file.isDirectory() ) continue;
+
+    Serial.printf("--- %s ---\n", file.name());
+
+    while ( file.available() )
+    {
+      uint32_t readlen;
+      char buffer[64] = { 0 };
+      readlen = file.read(buffer, sizeof(buffer));
+
+      Serial.print(buffer);
+    }
+    file.close();
+
+    Serial.println("---------------\n");
+    Serial.flush();
+  }
+}
+
+// the loop function runs over and over again forever
+void loop() 
+{
+  TaskHandle_t th = xTaskGetCurrentTaskHandle();
+
+  if ( millis() > TIME_OUT*1000 )
+  {
+    // low priority task print summary
+    if ( !summarized && (TASK_PRIO_LOW == uxTaskPriorityGet(th)))
+    {
+      summarized = true;
+      list_files();
+    }
+
+    vTaskSuspend(NULL); // suspend task
+    return;
+  }
+
+  const char* name = pcTaskGetName(th);
+  Serial.printf( "Task %s writing ...\n", name );
+  Serial.flush();
+
+  // Write files
+  write_files(name);
+
+  // lower delay increase chance for high prio task preempt others.
+  delay(100);
+}
--- a/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
+++ b/libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
@@ -15,6 +15,9 @@
 #include <Adafruit_LittleFS.h>
 #include <InternalFileSystem.h>
 
+// hardware random generator
+#include "nrf_rng.h"
+
 using namespace Adafruit_LittleFS_Namespace;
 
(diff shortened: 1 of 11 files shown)


hathach:
Thanks @henrygab for your superb works. Though I am currently in the middle of other works, need to wrap it up asap since I am blocking other people works. I will surely re-check/review all of your excellent work on LFS later on.

hathach:
Thanks @henrygab for your effort and patient :) . I am pull this PR to test locally with https://gist.github.com/henrygab/8fa35df56889d64b240e4f64d164f731 . I will do an overall review to this patch and the whole LFS as well. The current version used by the repo is 1.6, https://github.com/ARMmbed/littlefs LFS is currently at  2.1 . Maybe we could update it as well.

henrygab:
> ... The current version _[of LFS]_ used by the repo is 1.6, https://github.com/ARMmbed/littlefs LFS is currently at  2.1 . Maybe we could update it as well.

I would delay the LFS 1.6 to 2.x update for at least two reasons:

1. changing the cache layer at the same time as the file system major revision raises the risk of a bug (e.g., due to new requirements or new LFS behavior causing new interactions), and at the same time, if such a bug become real, it will be harder to know if the bug would have existed without the 2.0 LFS.  Changing one thing at a time helps track down problems.

2. LFS v1.x and LFS v2.x are [not compatible](https://github.com/ARMmbed/littlefs/pull/85/commits/4ad09d6c4ec95c126909b905900237cdc829c6b0).  Thus, some thought on how to handle this is recommended.  Do you include the migration code by default, to save existing media's data?  Or do you save 11% memory by ignoring that and forcing reformat / running an intermediate sketch that simply upgrades the FS version?  Or something else entirely?

It's of course your decision to make, I simply offer my thoughts based on past experiences.

Regardless, I'm thrilled to know this sporadic corruption bug might finally be fixed, and look forward to seeing this PR finally integrate, closing the book on this series of reported issues! 🎉

hathach:
> > ... The current version _[of LFS]_ used by the repo is 1.6, https://github.com/ARMmbed/littlefs LFS is currently at  2.1 . Maybe we could update it as well.
> 
> I would delay the LFS 1.6 to 2.x update for at least two reasons:
> 
> 1. changing the cache layer at the same time as the file system major revision raises the risk of a bug (e.g., due to new requirements or new LFS behavior causing new interactions), and at the same time, if such a bug become real, it will be harder to know if the bug would have existed without the 2.0 LFS.  Changing one thing at a time helps track down problems.
> 2. LFS v1.x and LFS v2.x are [not compatible](https://github.com/ARMmbed/littlefs/pull/85/commits/4ad09d6c4ec95c126909b905900237cdc829c6b0).  Thus, some thought on how to handle this is recommended.  Do you include the migration code by default, to save existing media's data?  Or do you save 11% memory by ignoring that and forcing reformat / running an intermediate sketch that simply upgrades the FS version?  Or something else entirely?
> 
> It's of course your decision to make, I simply offer my thoughts based on past experiences.
> 
> Regardless, I'm thrilled to know this sporadic corruption bug might finally be fixed, and look forward to seeing this PR finally integrate, closing the book on this series of reported issues!

Thanks for your opinion, I will make sure the stress test passed with current 1.6 version first. Then do some testing with 2.x . We will upgrade to 2.x (or any later version anyway). I am not really sure how long it takes until I could mange the time to come to this issue now. So I will try to see if I could get the best for this week :). 

I write a test sketch and submit it as PR to your repo folk https://github.com/henrygab/Adafruit_nRF52_Arduino/pull/2 , please check it out. 

As you recommendation, it uses 4 threads with different priority: loop (low), normal, high, highest. With small enough delay each run for task to preempt. It seems to lead to deadblock. I will do more tests.

https://gist.github.com/hathach/859d228204c97f130359b391d2f1ea76#file-internal_stresstest-ino

hathach:
forgot the log at Level 2

```
Internal Stress Test Example
Formatting ... Done
Task normal writing ...
Task high writing ...
Task normal writing ...
Task high writing ...
[IFLASH] Blocked parallel write attempt ... waiting for mutex
Task highest writing ...
Task high writing ...
Task highest writing ...
[IFLASH] Blocked parallel write attempt ... waiting for mutex
Task normal writing ...
Task highest writing ...
[IFLASH] Blocked parallel write attempt ... waiting for mutex
Task high writing ...
Task highest writing ...
[IFLASH] Bl
```

henrygab:
Based on my below comments, I am recommending to accept this PR as-is.

> forgot the log at Level 2
> [... snip!]
> [IFLASH] Blocked parallel write attempt ... waiting for mutex

This is great!  The fact that the output indicates blocked parallel write outputs validates that this fixes one cause of corruption!

I have not been able to reproduce a deadlock.  However, due to the output having an extremely high rate of bad block warnings, I dug deeper.  Then I discovered the following two notes:

* https://github.com/ARMmbed/littlefs/issues/124#issuecomment-457442048_
* https://github.com/ARMmbed/littlefs/issues/12#issuecomment-359612016_

To summarize... LFS *_is not_* re-entrant.  LFS is designed to only be used from a single thread / task at a time.  More specific rules:

* LFS allows multiple concurrent readers of a single file.
* LFS allows a single writer of a file (no concurrent readers allowed of that file)
* [conjecture] LFS allows multiple concurrent read-only metadata operations (directory enumeration, file open)
* [conjecture] LFS allows a single metadata modification operation (metadata reads must wait)

Put another way, it appears LFS has a shared-read / exclusive-write model.

If I understand correctly, the softdevice uses LFS, at least to store BLE bonding data.  Sketches also want to store data.  Thus, it appears that LFS entry points may need a mutex, similar to what was done here, if the goal is to allow multi-threaded concurrent LFS calls.

Even understanding the above, this PR prevents one layer of corruption, and it's unlikely that a sketch & the softdevice will both write to the same file.

henrygab:
Finally, I recommend delaying before updating to v2.x ... it appears they are similarly tracking a [power-failure corruption bug](https://github.com/ARMmbed/littlefs/issues/295) that is likely new to v2.x.

I'll open a new issue to cover the need to serialize LFS.

libraries/InternalFileSytem/examples/Internal_StressTest/Internal_StressTest.ino
```diff
@@ -0,0 +1,130 @@
...
+  Serial.println("Done"); Serial.flush();
+
+  // Create thread with different priority
+  // Although all the thread share loop() code, they are separated threads
+  // and running with different priorities
+
+  // Note: default loop() is running at LOW
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
+  Scheduler.startLoop(loop, 1024, TASK_PRIO_NORMAL, "normal");
```
hathach: I use threadname as filename for each thread to write, we should name them differently even though they share the priority. Maybe n1, n2, n3. I will made another PR   to update the sketch

hathach:
@henrygab I push another PR to your branch here 
 https://github.com/henrygab/Adafruit_nRF52_Arduino/pull/3 . It pretty reliable now when having each thread writing to its own directory. I run for mini test for 20 seconds, and 100ms delay. There is no bad/corrupted blocks as previously, although there is bunch of `waiting for mutex`. Possibly concurrently writing to the root folder make it corrupted. 

here is the log, there is 66 writes, and they are all visible from thread's file, not one missing. 
https://gist.github.com/hathach/7c3fad9b2119b90e5acd99175ac0b6fe
 
This looks good enough, I will let the sketch running for longer period to see if there is any issues. Else I think we can safely merge this. Current scenario Bluefruit lib using LFS to store bond in its own subfolder "/adafruit/bond_prph" and "/adafruit/bond_cntr", and this isn't always read/write. It should relatively safe to use together with user sketch, unless yeah for power corruption issue.

hathach:
After lots of testing, there is still issue running the stresstest over long period of time (600 seconds) with 2000 writes occasionally. Though, it could be due to lfs internal issue, I am not sure. But it is much more reliable than ever. Thank you very much @henrygab  for your great work.

libraries/InternalFileSytem/src/flash/flash_cache.c
```diff
@@ -22,12 +22,69 @@
...
 
+
+
+static volatile SemaphoreHandle_t _serializeFlashAccess = NULL;
+
+int  _internal_flash_cache_write (flash_cache_t* fc, uint32_t dst, void const * src, uint32_t len);
+int  _internal_flash_cache_read  (flash_cache_t* fc, void* dst, uint32_t addr, uint32_t count);
+void _internal_flash_cache_flush (flash_cache_t* fc);
+
+static inline void _internal_EnsureFlashCacheSemaphoreInitialized()
```
hathach: I am doing mostly testing and tweaking so far. Everything is good, though this seems to be over complicated. I will add an Static semaphor struct, it will guarantee an mutex creation is always succeeded in my PR.

henrygab:
> After lots of testing, there is still issue running the stresstest over long period of time (600 seconds) with 2000 writes occasionally. Though, it could be due to lfs internal issue, I am not sure. But it is much more reliable than ever. Thank you very much @henrygab  for your great work.

Happy to see it working well.  BTW, the reason it's now stable is not so much because the threads were all writing to the **_root_** directory per se, but rather that they were all writing to the **_same_** directory.  At least in LFS 1.6 (and FAT12, FAT16, FAT32, and exFAT), the metadata for each directory tends to be in its own cluster / sector / allocation unit.  While there are still shared metadata updates when creating the directories in the root, and when allocating additional pages when the file grows to a new page, the test sketch is no longer likely to cause LFS to simultaneously attempt to update the same internal state, as each file has its own cluster to play in.

There is absolutely still a lingering potential problem if two tasks update the file system, as LFS itself is not serialized.  However, it seems if a sketch can apply the following guidance, it might reduce risk of corruption in LFS itself:

1. Before initializing BT LE or creating any tasks that would use the file system:
    a. Initialize the internal file system.
    b. Create a distinct subdirectory for each task that might simultaneously use the filesystem.
    c. Where possible, pre-allocate the space for each file (e.g., seek to size needed and write a single byte)
    d. Flush() to ensure it's all stored on the flash
2. Then, only after the subdirectories are created and files pre-allocated, initialize BT LE

This should further improve reliability, until (if) the LFS upper layer can be serialized also.

hathach:
@henrygab  https://github.com/henrygab/Adafruit_nRF52_Arduino/pull/4 use static semaphore create which is always succedded.  This PR  is as good as I could hope, then I think it makes sense to move the mutex into `flash_cache_t` to make the code more generic. Afterwards, I see the lack of init() is a bit troublesome, making the API checking semaphore every time is a waste of cycle --> flash cache init. Please let me know what do you think of the changes, I am open to any suggestion.

henrygab:
Will look soon!

hathach:
Take your time, I am always having overdue work to tackle 🤓🤓

henrygab:
@hathach -- I realized you indicated this flash cache layer may also be used in other project.  If that is true, then even though these changes will not be necessary for the InternalFS on nRF52, perhaps it would be worthwhile to merge these changes anyways, in case any other AdaFruit project wants to use this flash cache layer where it's **_not_** guaranteed that higher-level code will synchronize?

What are your thoughts on that?

henrygab:
@hathach -- If you use this caching layer for any other adafruit project, I recommend you take these changes, in case other projects do not synchronize the cache layer.  Otherwise, it may be OK to close this one entirely now.

hathach:
@henrygab thank for reminding, close this for now

PR #377, opened 2019-11-08
Reduce firewall to minimum

henrygab:
That is, block only traffic incoming to 5353/udp and traffic targeting the mDNS multicast address.


henrygab:
just a bit early... and I'll be squash merging into the prior PR.

Issue #72, opened 2017-08-01
pull request to firmata for our nrf52 Boards.h 


hathach:
Currently firmata library is included as part of the repo since official repo Boads.h does not include our definition yet.
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/libraries/Firmata/Boards.h

After testing it reliably, we should submit a pull request to got official support from the library.
#58

hathach:
pull request 
https://github.com/firmata/arduino/pull/379

PR #397, opened 2019-12-20
Serialize lfs

### LINKED ISSUE #222: Connecting/Reconnecting a nrf52 peripheral to a central can corrupt the LittleFS FS where pairing information is stored.

### LINKED ISSUE #227: InternalFS Corrupt Blocks

### LINKED ISSUE #325: nRF52832 frequently connect/disconnect occur assertion "head >= 2 && head <= lfs->cfg->block_count"

### LINKED ISSUE #350: [Bug] InternalFS analysis: LFS requirements, coherency guarantees

### LINKED ISSUE #393: LFS needs to be synchronized
**Describe the bug**
LFS requires synchronization if used from multiple threads.

  - nRF52840 Feather Express, 0.14.6 + commit aed854c1069050d5aa5c2821eb4c9fb744d240dc
  - Log and sketch at https://gist.github.com/henrygab/b7d5376a09e8a488483347fc100a1ba4

**To Reproduce**
1. Build the sketch and upload to supported nRF52840 board
2. Watch log

**Expected behavior**
Few (if any) bad blocks detected

**Additional context**
Turns out that LFS requires, when multiple threads may use it simultaneously, that locking be added (e.g., with mutex).

See partial fix via issue #350 and PR #372.
See [questions](https://github.com/ARMmbed/littlefs/issues/352) posted at littlefs repo.

henrygab:
Fix #350
Fix #325
Fix #227
Fix #222
Fix #393

@hathach -- This PR supercedes #372 (Guard against race conditions in Flash FS cache), as #72 was not really providing a proper fix of the underlying problem:

LFS itself is [not safe](https://github.com/ARMmbed/littlefs/issues/352) to call simultaneously from multiple contexts.

This PR solves this by wrapping the InternalFS API with a non-recursive mutex, as recommended.

I realize the PR uses newly declared wrapper functions.  I tried to avoid this, but the changes got messy quite quickly, mostly due to some of the macros used to verify return codes exiting the function (which would leak the mutex).  This is a reasonably clean solution, without the more drastic move to use of RAII objects such as [`std::lock_guard<>`](https://en.cppreference.com/w/cpp/thread/lock_guard).

### DIFF
--- a/cores/nRF5/rtos.cpp
+++ b/cores/nRF5/rtos.cpp
@@ -45,13 +45,12 @@ static void _redirect_task(void* arg)
   while(1)
   {
     taskfunc();
-    delay(0);
+    yield();
   }
 }
 
 SchedulerRTOS::SchedulerRTOS(void)
 {
-  _num = 1; // loop is already created by default
 }
 
 bool SchedulerRTOS::startLoop(taskfunc_t task, uint32_t stack_size, uint32_t prio, const char* name)
--- a/cores/nRF5/rtos.cpp
+++ b/cores/nRF5/rtos.cpp
@@ -45,6 +45,7 @@ static void _redirect_task(void* arg)
   while(1)
   {
     taskfunc();
+    delay(0);
   }
 }
 
@@ -53,25 +54,15 @@ SchedulerRTOS::SchedulerRTOS(void)
   _num = 1; // loop is already created by default
 }
 
-bool SchedulerRTOS::startLoop(taskfunc_t task, uint32_t stack_size)
+bool SchedulerRTOS::startLoop(taskfunc_t task, uint32_t stack_size, uint32_t prio, const char* name)
 {
-  char name[8] = "loop0";
-  name[4] += _num;
-
-  if ( startLoop(task, name, stack_size) )
+  static char const * const name_default = "unnamed_task";
+  if (name == NULL)
   {
-    _num++;
-    return true;
-  }else
-  {
-    return false;
+    name = name_default;
   }
-}
-
-bool SchedulerRTOS::startLoop(taskfunc_t task, const char* name, uint32_t stack_size)
-{
   TaskHandle_t  handle;
-  return pdPASS == xTaskCreate( _redirect_task, name, stack_size, (void*) task, TASK_PRIO_LOW, &handle);
+  return pdPASS == xTaskCreate( _redirect_task, name, stack_size, (void*) task, prio, &handle);
 }
 
 
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
@@ -121,10 +121,8 @@ bool Adafruit_LittleFS::format (void)
 // Open a file or folder
 Adafruit_LittleFS_Namespace::File Adafruit_LittleFS::open (char const *filepath, uint8_t mode)
 {
-  xSemaphoreTake(_mutex,  portMAX_DELAY);
-  Adafruit_LittleFS_Namespace::File ret = Adafruit_LittleFS_Namespace::File(filepath, mode, *this);
-  xSemaphoreGive(_mutex);
-  return ret;
+  // No lock is required here ... the File() object will synchronize with the mutex provided
+  return Adafruit_LittleFS_Namespace::File(filepath, mode, *this);
 }
 
 // Check if file or folder exists
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
@@ -44,7 +44,7 @@ Adafruit_LittleFS::Adafruit_LittleFS (struct lfs_config* cfg)
   varclr(&_lfs);
   _lfs_cfg = cfg;
   _mounted = false;
-  _mutex = xSemaphoreCreateMutexStatic(&this->xMutexStorageSpace);
+  _mutex = xSemaphoreCreateMutexStatic(&this->_MutexStorageSpace);
 }
 
 Adafruit_LittleFS::~Adafruit_LittleFS ()
@@ -57,176 +57,159 @@ Adafruit_LittleFS::~Adafruit_LittleFS ()
 // User should format the disk and try again
 bool Adafruit_LittleFS::begin (struct lfs_config * cfg)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_begin(cfg);
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  bool ret;
+  // not a loop, just an quick way to short-circuit on error
+  do {
+    if (_mounted) { ret = true; break; }
+    if (cfg) { _lfs_cfg = cfg; }
+    if (nullptr == _lfs_cfg) { ret = false; break; }
+    // actually attempt to mount, and log error if one occurs
+    int err = lfs_mount(&_lfs, _lfs_cfg);
+    PRINT_LFS_ERR(err);
+    _mounted = (err == LFS_ERR_OK);
+    ret = _mounted;
+  } while(0);
   xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_begin (struct lfs_config * cfg)
-{
-  if ( _mounted ) return true;
-
-  if (cfg) _lfs_cfg = cfg;
-  if (!_lfs_cfg) return false;
-
-  VERIFY_LFS(lfs_mount(&_lfs, _lfs_cfg), false);
-  _mounted = true;
-
-  return true;
+  return ret;
 }
 
 // Tear down and unmount file system
 void Adafruit_LittleFS::end(void)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  this->xWrap_end();
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  if (_mounted)
+  {
+    _mounted = false;
+    int err = lfs_unmount(&_lfs);
+    PRINT_LFS_ERR(err);
+    (void)err;
+  }
   xSemaphoreGive(_mutex);
-}
-
-void Adafruit_LittleFS::xWrap_end(void)
-{
-  if (!_mounted) return;
-
-  _mounted = false;
-  VERIFY_LFS(lfs_unmount(&_lfs), );
 }
 
 bool Adafruit_LittleFS::format (void)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_format();
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  int err = LFS_ERR_OK;
+  bool attemptMount = _mounted;
+  // not a loop, just an quick way to short-circuit on error
+  do
+  {
+    // if already mounted: umount first -> format -> remount
+    if (_mounted)
+    {
+      _mounted = false;
+      err = lfs_unmount(&_lfs);
+      if ( LFS_ERR_OK != err) { PRINT_LFS_ERR(err); break; }
+    }
+    err = lfs_format(&_lfs, _lfs_cfg);
+    if ( LFS_ERR_OK != err ) { PRINT_LFS_ERR(err); break; }
+
+    if (attemptMount)
+    {
+      err = lfs_mount(&_lfs, _lfs_cfg);
+      if ( LFS_ERR_OK != err ) { PRINT_LFS_ERR(err); break; }
+      _mounted = true;
+    }
+    // success!
+  } while(0);
   xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_format (void)
-{
-  // if already mounted: umount -> format -> remount
-  if(_mounted) VERIFY_LFS(lfs_unmount(&_lfs), false);
-
-  VERIFY_LFS(lfs_format(&_lfs, _lfs_cfg), false);
-
-  if (_mounted) VERIFY_LFS(lfs_mount(&_lfs, _lfs_cfg), false);
-
-  return true;
+  return LFS_ERR_OK == err;
 }
 
 // Open a file or folder
 Adafruit_LittleFS_Namespace::File Adafruit_LittleFS::open (char const *filepath, uint8_t mode)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  Adafruit_LittleFS_Namespace::File retval = this->xWrap_open(filepath, mode);
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  Adafruit_LittleFS_Namespace::File ret = Adafruit_LittleFS_Namespace::File(filepath, mode, *this);
   xSemaphoreGive(_mutex);
-  return retval;
-}
-
-Adafruit_LittleFS_Namespace::File Adafruit_LittleFS::xWrap_open (char const *filepath, uint8_t mode)
-{
-  return Adafruit_LittleFS_Namespace::File(filepath, mode, *this);
+  return ret;
 }
 
 // Check if file or folder exists
 bool Adafruit_LittleFS::exists (char const *filepath)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_exists(filepath);
+  struct lfs_info info;
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  bool ret = (0 == lfs_stat(&_lfs, filepath, &info));
   xSemaphoreGive(_mutex);
-  return retval;
+  return ret;
 }
 
-bool Adafruit_LittleFS::xWrap_exists (char const *filepath)
-{
-  struct lfs_info info;
-  return 0 == lfs_stat(&_lfs, filepath, &info);
-}
 
 // Create a directory, create intermediate parent if needed
 bool Adafruit_LittleFS::mkdir (char const *filepath)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_mkdir(filepath);
-  xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_mkdir (char const *filepath)
-{
+  bool ret = true;
   const char* slash = filepath;
   if ( slash[0] == '/' ) slash++;    // skip root '/'
 
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+
+  // make intermediate parent directory(ies)
   while ( NULL != (slash = strchr(slash, '/')) )
   {
     char parent[slash - filepath + 1] = { 0 };
     memcpy(parent, filepath, slash - filepath);
 
-    // make intermediate parent
     int rc = lfs_mkdir(&_lfs, parent);
     if ( rc != LFS_ERR_OK && rc != LFS_ERR_EXIST )
     {
       PRINT_LFS_ERR(rc);
-      return false;
+      ret = false;
+      break;
     }
-
     slash++;
   }
-  
-  int rc = lfs_mkdir(&_lfs, filepath);
-  if ( rc != LFS_ERR_OK && rc != LFS_ERR_EXIST )
+  // make the final requested directory
+  if (ret)
   {
-    PRINT_LFS_ERR(rc);
-    return false;
+    int rc = lfs_mkdir(&_lfs, filepath);
+    if ( rc != LFS_ERR_OK && rc != LFS_ERR_EXIST )
+    {
+      PRINT_LFS_ERR(rc);
+      ret = false;
+    }
   }
-
-  return true;
+  xSemaphoreGive(_mutex);
+  return ret;
 }
 
 // Remove a file
 bool Adafruit_LittleFS::remove (char const *filepath)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_remove(filepath);
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  int err = lfs_remove(&_lfs, filepath);
+  PRINT_LFS_ERR(err);
   xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_remove (char const *filepath)
-{
-  VERIFY_LFS(lfs_remove(&_lfs, filepath), false);
-  return true;
+  return LFS_ERR_OK == err;
 }
 
 // Remove a folder
 bool Adafruit_LittleFS::rmdir (char const *filepath)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_rmdir(filepath);
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  int err = lfs_remove(&_lfs, filepath);
+  PRINT_LFS_ERR(err);
   xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_rmdir (char const *filepath)
-{
-  VERIFY_LFS(lfs_remove(&_lfs, filepath));
-  return true;
+  return LFS_ERR_OK == err;
 }
 
 // Remove a folder recursively
 bool Adafruit_LittleFS::rmdir_r (char const *filepath)
 {
-  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
-  bool retval = this->xWrap_rmdir_r(filepath);
-  xSemaphoreGive(_mutex);
-  return retval;
-}
-
-bool Adafruit_LittleFS::xWrap_rmdir_r (char const *filepath)
-{
   /* adafruit: lfs is modified to remove non-empty folder,
    According to below issue, comment these 2 line won't corrupt filesystem
-   https://github.com/ARMmbed/littlefs/issues/43 */
-  VERIFY_LFS(lfs_remove(&_lfs, filepath));
-  return true;
+   at least when using LFS v1.  If moving to LFS v2, see tracked issue
+   to see if issues (such as the orphans in threaded linked list) are resolved.
+   https://github.com/ARMmbed/littlefs/issues/43
+   */
+  xSemaphoreTake(_mutex,  portMAX_DELAY);
+  int err = lfs_remove(&_lfs, filepath);
+  PRINT_LFS_ERR(err);
+  xSemaphoreGive(_mutex);
+  return LFS_ERR_OK == err;
 }
 
 //------------- Debug -------------//
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
@@ -44,6 +44,7 @@ Adafruit_LittleFS::Adafruit_LittleFS (struct lfs_config* cfg)
   varclr(&_lfs);
   _lfs_cfg = cfg;
   _mounted = false;
+  _mutex = xSemaphoreCreateMutexStatic(&this->xMutexStorageSpace);
 }
 
 Adafruit_LittleFS::~Adafruit_LittleFS ()
@@ -55,6 +56,14 @@ Adafruit_LittleFS::~Adafruit_LittleFS ()
 // Return true if mounted successfully else probably corrupted.
 // User should format the disk and try again
 bool Adafruit_LittleFS::begin (struct lfs_config * cfg)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_begin(cfg);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_begin (struct lfs_config * cfg)
 {
   if ( _mounted ) return true;
 
@@ -70,6 +79,13 @@ bool Adafruit_LittleFS::begin (struct lfs_config * cfg)
 // Tear down and unmount file system
 void Adafruit_LittleFS::end(void)
 {
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  this->xWrap_end();
+  xSemaphoreGive(_mutex);
+}
+
+void Adafruit_LittleFS::xWrap_end(void)
+{
   if (!_mounted) return;
 
   _mounted = false;
@@ -77,6 +93,14 @@ void Adafruit_LittleFS::end(void)
 }
 
 bool Adafruit_LittleFS::format (void)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_format();
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_format (void)
 {
   // if already mounted: umount -> format -> remount
   if(_mounted) VERIFY_LFS(lfs_unmount(&_lfs), false);
@@ -91,11 +115,27 @@ bool Adafruit_LittleFS::format (void)
 // Open a file or folder
 Adafruit_LittleFS_Namespace::File Adafruit_LittleFS::open (char const *filepath, uint8_t mode)
 {
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  Adafruit_LittleFS_Namespace::File retval = this->xWrap_open(filepath, mode);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+Adafruit_LittleFS_Namespace::File Adafruit_LittleFS::xWrap_open (char const *filepath, uint8_t mode)
+{
   return Adafruit_LittleFS_Namespace::File(filepath, mode, *this);
 }
 
 // Check if file or folder exists
 bool Adafruit_LittleFS::exists (char const *filepath)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_exists(filepath);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_exists (char const *filepath)
 {
   struct lfs_info info;
   return 0 == lfs_stat(&_lfs, filepath, &info);
@@ -103,6 +143,14 @@ bool Adafruit_LittleFS::exists (char const *filepath)
 
 // Create a directory, create intermediate parent if needed
 bool Adafruit_LittleFS::mkdir (char const *filepath)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_mkdir(filepath);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_mkdir (char const *filepath)
 {
   const char* slash = filepath;
   if ( slash[0] == '/' ) slash++;    // skip root '/'
@@ -136,6 +184,14 @@ bool Adafruit_LittleFS::mkdir (char const *filepath)
 // Remove a file
 bool Adafruit_LittleFS::remove (char const *filepath)
 {
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_remove(filepath);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_remove (char const *filepath)
+{
   VERIFY_LFS(lfs_remove(&_lfs, filepath), false);
   return true;
 }
@@ -143,12 +199,28 @@ bool Adafruit_LittleFS::remove (char const *filepath)
 // Remove a folder
 bool Adafruit_LittleFS::rmdir (char const *filepath)
 {
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_rmdir(filepath);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_rmdir (char const *filepath)
+{
   VERIFY_LFS(lfs_remove(&_lfs, filepath));
   return true;
 }
 
 // Remove a folder recursively
 bool Adafruit_LittleFS::rmdir_r (char const *filepath)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
+  bool retval = this->xWrap_rmdir_r(filepath);
+  xSemaphoreGive(_mutex);
+  return retval;
+}
+
+bool Adafruit_LittleFS::xWrap_rmdir_r (char const *filepath)
 {
   /* adafruit: lfs is modified to remove non-empty folder,
    According to below issue, comment these 2 line won't corrupt filesystem
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
@@ -46,8 +46,19 @@ File::File (Adafruit_LittleFS &fs)
 File::File (char const *filename, uint8_t mode, Adafruit_LittleFS &fs)
  : File(fs)
 {
+  // public constructor calls public API open(), which will obtain the mutex
   this->open(filename, mode);
 }
+
+File::File (uint8_t differentiator, char const *filename, uint8_t mode, Adafruit_LittleFS &fs)
+ : File(fs)
+{
+  // private constructor calls private API _open(), which does NOT obtain the mutex
+  (void)differentiator;
+  this->_open(filename, mode);
+}
+
+
 
 bool File::_open_file (char const *filepath, uint8_t mode)
 {
@@ -99,12 +110,23 @@ bool File::_open_dir (char const *filepath)
   return true;
 }
 
+// TODO: wrap close() because it must be called by open()
+// TODO: wrap open() because it is called by openNextFile()
+
 bool File::open (char const *filepath, uint8_t mode)
+{
+  bool ret = false;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
+  ret = this->_open(filepath, mode);
+  this->DoNotCallFromOutsideClass_UnlockFilesystem();
+  return ret;
+}
+bool File::_open (char const *filepath, uint8_t mode)
 {
   bool ret = false;
 
   // close if currently opened
-  if ( (*this) ) close();
+  if ( this->_isOpen() ) _close();
 
   struct lfs_info info;
   int rc = lfs_stat(_fs->getFS(), filepath, &info);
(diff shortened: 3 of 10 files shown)


mattratto:
This is great - thanks Henry and all. I have seen a file corruption issue remaining with 372 (much more infrequent than before but still there). Wonderful to have this wrapped up.

henrygab:
@mattratto  -- You're right!  This will be wonderful to have wrapped up.

It'd be great to have some indication that this resolves a real-world reported instance, such as your remaining corruption problem.  Would you be willing to reply confirming that you no longer see your remaining corruption after using this fix?  (after you are fairly confident, of course)

Let's hope this closes this one for good!

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
```diff
@@ -55,6 +56,14 @@ Adafruit_LittleFS::~Adafruit_LittleFS ()
 // Return true if mounted successfully else probably corrupted.
 // User should format the disk and try again
 bool Adafruit_LittleFS::begin (struct lfs_config * cfg)
+{
+  while (pdTRUE != xSemaphoreTake(_mutex,  portMAX_DELAY)) {}
```
hathach: I think we don't need to check the return with `portMAX_DELAY` since it literally means wait forever.

hathach:
@henrygab  do you think we also need mutex for the `Adafruit_LittleFS_File` as well ?
 https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp

hathach:
I make PR to your branch to unwrap the VERIFY_LFS() here, since it it a bit too much work to ask you to change, please let's me know what you think. 
https://github.com/henrygab/Adafruit_nRF52_Arduino/pull/5

hathach:
I still see the assert file with the stress test, we definitely need to wrap up the `Adafruit_LittleFS_File` as well

```
Task n3 writing ...
Task loop writing ...
Task loop writing ...
assertion "head >= 2 && head <= lfs->cfg->block_count" failed: file "/home/hathach/Arduino/hardware/adafruit/nrf52/libraries/Adafruit_LittleFS/src/littlefs/lfs.c", line 1235, function: lfs_ctz_extend
Task high writing ...
Task n3 writing ...
Task n1 writing ...
Task n3 writing ...
```

cores/nRF5/rtos.cpp
```diff
@@ -45,6 +45,7 @@ static void _redirect_task(void* arg)
   while(1)
   {
     taskfunc();
+    delay(0);
```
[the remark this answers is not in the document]
henrygab: ok.  For some reason, when I traced through the yield() earlier, I thought it failed to call `tud_cdc_write_flush()` … but that is not the case … it was my error.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS.cpp
henrygab: I did not realize that there were no other error conditions here.  You are correct again.

henrygab:
> @henrygab  do you think we also need mutex for the Adafruit_LittleFS_File as well ?

Yes, I think we need to also use the mutex when lfs APIs are used from within `Adafruit_LittleFS_File`.
Let me see if I can create a small patch, likely by declaring a friend function (or two).

henrygab:
@hathach -- the `Adafruit_LittleFS_File` public APIs appear to call into other public APIs, and thus would deadlock if simply ensuring the lock is acquired at each public API function entry / released at each exit.

I understand that you have indicated that you do not prefer the use of a wrapper function.  I am happy to consider alternatives, but believe the safest change is to wrap each APIs that is called from  another public API, so as to allow the serialization without major changes to the existing, tested, code flow.

What do you think?

henrygab:
@hathach To maximize working time, I've gone ahead and put forward change that continues to work for the stress test.

It was not as simple as I had hoped.  Two issues in particular arose:

1. Because the `Adafruit_LittleFS_File` constructors were public, it was not possible to modify them without breaking the existing API.  Although I preferred otherwise, I added some friendship between the classes.

2. The implementation of various `Adafruit_LittleFS_File` functions called back into its own public API.  This meant that either a recursive mutex would be required (not a great solution), or that some of those public functions would need convert to wrappers around private functions.  I chose the latter solution of (limited) wrapper functions where needed.

3. As before, the use of the VERIFY() macros had to be unwound.  This raises the possibility of typos or logic inversion.  

Accordingly, I strongly request your careful code review and testing for commit 2a2f771.

henrygab:
Rebase changes to be based from 0.15.0 ...

mattratto:
I haven’t tried the rebase, but use of the previous version still results in periodic lockups - only on sequential removal of files. This is happening after the saving of approximately 100 separate 50byte files. When I loop through and remove them, sometimes the code halts. I haven’t yet run this with debug on, but once I do, I will post more info. 

I am wondering if I am not allowing enough of a delay between each remove, though from my reading of the new code this should not be an issue. 

I should add that this is occurring in a production environment which includes Bluetooth transactions. 

Matt

mattratto:
Another note - this is occurring on brand new boards. I am not running a file system format before placing files on the flash, following initial install of software. Is it possible that reliability would be increased by adding an initial format prior to adding files?

hathach:
Sorry for late response, is busy migrating from travis to github action, and bunch o f unnamed work at the end of year. Will look at this next.

henrygab:
@mattratto -- 

Question: Are your files in the root directory, or are they all in a subdirectory that only you use?

**_Why:_** 

As a test, if you they were in root, and you moved all your files out into a subdirectory, would you still see a problem?

* If yes, then I'm intrigued... Could the softdevice also doing some access to the flash that fundamentally conflicts? Could it be that it cannot be "safe" to cache data at all when the softdevice is enabled?
* If no, then while not definitive (cannot easily tell difference between move from 99% --> 99.999% vs. to 100%), using the subdirectory may be a quick way to reduce instances of the problem?

mattratto:
They were all in root. I’m moving to subdirectory and continuing to test. Is there anything fancy I should be doing with this? I’ve just appended /data/ to the file names for the write, and open /data/ for the reads and deletes. 

As always, my thanks Henry!

henrygab:
_First, I have very little understanding of what the SD does.  This currently is limiting me to (hopefully educated) guesses on how this might occur.  That said..._

No, nothing fancy.  A subdirectory is all that's needed to get the potential reduction in interactions.

@hathach -- do you know when, why, and how the SD is saving bluetooth pairing data to flash?

mattratto:
So I’ve put my data files in a directory and now, instead of deleting them individually, I delete and recreate the directory. Either the use of the directory, or this new deletion process seems to have resolved remaining issues. I am of course continuing to test, but looking good so far!

henrygab:
@mattratto That data supports my current suspicion that the softdevice is messing with (writing directly to) sectors that Adafruit_LittleFS is using.

_(( I'm curious if there is any repro without sd enabled, but I can no longer repro the issue with this branch ))_

Can you get a repro if you go back to the prior behavior (saving in root directory), and *not* enable the softdevice / bluetooth?  Obviously, this presumes you have multiple tasks that each use the filesystem.

If you can repro even without the softdevice enabled, then something else is awry.

henrygab:
@mattratto -- Matt, I want to be clear that, if the SD is the cause of corruption by writing to the LFS file system without being serialized, using a subdirectory will only make the problem much more rare, and not eliminate it entirely.

This is because there are still globally shared data structures on the media, so if/when certain conditions get hit, it would be more likely to cause conflict.

I think your prior method of deleting many files in sequence may have been a good test case _(but also good to know the recursive directory deletion can also further reduce the likelihood of conflict with SD)_.

hathach:
Sorry for another delay, was busy with other works lately. 

> @hathach -- do you know when, why, and how the SD is saving bluetooth pairing data to flash?

To be clear SD doesn't write to LFS, only the Bluefruit library save information to LFS when
- Paired with an new device
- CCCD of bonded device is changed (enabled or disabled)

CCCD changes occurs more often then pairing with new device, if there is conflict, it is probably it.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
```diff
@@ -69,13 +75,21 @@ class File : public Stream
...
+
+    bool isOpen(void);
     char const* name (void);
 
     bool isDirectory (void);
     File openNextFile (uint8_t mode = FILE_O_READ);
     void rewindDirectory (void);
 
+    // Has to be public, in order to grant 'friend' permissions to Adafruit_LittleFS private member _mutex
+    void DoNotCallFromOutsideClass_LockFilesystem(void);
```
hathach: @henrygab  Do we need to lock the entire file system when working with File API, I am thinking we could have an mutex for each file/folder to ensure that there are 2 threads that open the same file/directory instead of locking the whole file system.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
```diff
@@ -69,13 +75,21 @@ class File : public Stream
     uint32_t size (void);
 
     void close (void);
+
+    __attribute__((deprecated("Recommend use of isOpen() for clarity")))
```
hathach: it is how Arduino design and used to test the file through out its examples. Don't be bothered with it ;)

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
```diff
@@ -69,13 +75,21 @@ class File : public Stream
...
     operator bool (void);
+
+    bool isOpen(void);
     char const* name (void);
 
     bool isDirectory (void);
     File openNextFile (uint8_t mode = FILE_O_READ);
     void rewindDirectory (void);
 
+    // Has to be public, in order to grant 'friend' permissions to Adafruit_LittleFS private member _mutex
```
hathach: @henrygab as long as it works, maybe we should expose the LFS lock/unlock() file system to public :)

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -46,9 +46,20 @@ File::File (Adafruit_LittleFS &fs)
...
 {
+  // public constructor calls public API open(), which will obtain the mutex
   this->open(filename, mode);
 }
 
+File::File (uint8_t differentiator, char const *filename, uint8_t mode, Adafruit_LittleFS &fs)
+ : File(fs)
+{
+  // private constructor calls private API _open(), which does NOT obtain the mutex
+  (void)differentiator;
```
hathach: could you tell me what is the `differentiator` for ?

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
 
 size_t File::write (uint8_t const *buf, size_t size)
 {
-  VERIFY(!_is_dir, 0);
-
-  lfs_ssize_t wrcount = lfs_file_write(_fs->getFS(), _file, buf, size);
-  VERIFY(wrcount > 0, 0);
+  lfs_ssize_t wrcount = 0;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
+  if (!this->_is_dir)
```
hathach: this should be checked before attempting to lock the fs

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
+    };
+  }
+  this->DoNotCallFromOutsideClass_UnlockFilesystem();
   return wrcount;
 }
 
 int File::read (void)
 {
-  VERIFY(!_is_dir, -1);
+  // this thin wrapper relies on called function to synchronize
```
hathach: _is_dir shouldn't be dropped

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
       _file = NULL;
     }
   }
 }
 
+[[deprecated("Recommend use of isOpen() for clarity")]]
 File::operator bool (void)
+{
+  bool ret = false;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
```
hathach: there is no need to lock fs for this function. We can also make bool() as inline verion of isOpen()

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
 
+// WARNING -- although marked as `const`, the values pointed
+//            to may change.  For example, if the same File
+//            object has `open()` called with a different
+//            file or directory name, this same pointer will
+//            suddenly (unexpectedly?) have different values.
 char const* File::name (void)
 {
-  return _name;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
```
hathach: I don't see why we have to lock fs for these API(), it just a simple return.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
+  char const* ret = this->_name;
+  this->DoNotCallFromOutsideClass_UnlockFilesystem();
+  return ret;
 }
 
 bool File::isDirectory (void)
 {
-  return _is_dir;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
```
hathach: same as above

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
+}
+bool File::isOpen(void)
+{
+  bool ret = 0;
+  this->DoNotCallFromOutsideClass_LockFilesystem();
+  ret = this->_isOpen();
+  this->DoNotCallFromOutsideClass_UnlockFilesystem();
+  return ret;
+}
+bool File::_isOpen(void)
```
hathach: I think we don't need _isOpen() variant, just put this implementation as isOpen()

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -130,151 +152,260 @@ bool File::open (char const *filepath, uint8_t mode)
...
+  this->DoNotCallFromOutsideClass_LockFilesystem();
+  if (!this->_is_dir)
+  {
+    ret = lfs_file_read(_fs->getFS(), _file, buf, nbyte);
+  }
+  this->DoNotCallFromOutsideClass_UnlockFilesystem();
+  return ret;
 }
 
 int File::peek (void)
```
hathach: I think peek() is safe as it is

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: The `_isOpen()` is necessary still because this function was used in other internal APIs, while the mutex was already taken.  In order to keep the same code flow and readability, this API is necessary to have as an internal function.  See, for example, `_open()` and `_close()`.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: The member variable `_is_dir` can be modified by a call to `File::open( … )`.  Therefore, access to the function must be serialized.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: The member variable `_name` can be modified by a call to `File::open( … )`.  Therefore, access to the function must be serialized.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: OK.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: I disagree.  The same `File` object can change from a file to a directory, such as through a call to `File::open(char const *filepath, uint8_t mode)`.  Therefore, it is necessary to lock before checking the member variable `_is_dir`.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
henrygab: Good question.

This was a relic of a prior attempted solution, before `open()` was split into `open()` as a wrapper and `_open()` as the internal function without mutex.

<detail>
<summary>Background, in case you wonder...</summary>

Let me give some background on existing constructors and use:

* `File::File (Adafruit_LittleFS &fs)` -- This public constructor simply keeps a copy of the pointer.  No lock used.

* `File::File (char const *filename, uint8_t mode, Adafruit_LittleFS &fs) : File(fs)` -- This calls the above constructor, but then calls `open()`, which will obtain the mutex as part of the open().

At the time, I believe `openNextFile()` created the return value by using the second constructor (with the filepath + mode).  Therefore, it was necessary for the mutex to **NOT** be acquired to call that constructor, to avoid deadlock.  Obviously, this was not OK, so I added the private constructor with a differentiating argument, so it would only be callable from within the `File` class.

</detail>

But, as your sharp eyes noticed, this constructor is no longer necessary.  I've removed it now.

henrygab: Background, in case you wonder...

Let me give some background on existing constructors and use:

* `File::File (Adafruit_LittleFS &fs)` -- This public constructor simply keeps a copy of the pointer.  No lock used.

* `File::File (char const *filename, uint8_t mode, Adafruit_LittleFS &fs) : File(fs)` -- This calls the above constructor, but then calls `open()`, which will obtain the mutex as part of the open().

At the time, I believe `openNextFile()` created the return value by using the second constructor (with the filepath + mode).  Therefore, it was necessary for the mutex to **NOT** be acquired to call that constructor, to avoid deadlock.  Obviously, this was not OK, so I added the private constructor with a differentiating argument, so it would only be callable from within the `File` class.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
henrygab: Ok.

Because this PR fixes data corruption, I think it is important to accept even if you want to change some function names later.

Separately, if you want the design to better hide this, I would recommend following the PIMPL (Pointer-to-Implementation) solution.  This would allow 100% hiding of all the non-public APIs, and the interactions between the FS / File abstractions, without exposing any of this internal shared locking.  However, changing to PIMPL would be a bigger architectural change … my goal was to minimize changes while serializing access to LFS APIs.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
henrygab: Ok...

hathach:
hmm, look like you are too quick, I made a modification and submit it as PR (as usual) to your branch, but look like github has difficulty to pick up the differences.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
hathach: IMHO, it is over-engineer,  I don't see use case where user change want to change it to other filepath in another thread out of context.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
hathach: I mean there is no need to lock mutex when checking with _file and _dir member. They should be reachable all time.

libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.h
hathach: No, I don't need any fancy way to absolute hide API from user. I would simply add underscore to `_lock/_unlock` public API  with comment saying it is intended to be used internally  and expect user not to call it :D
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/Bluefruit52Lib/src/bluefruit.h#L239

hathach:
Thank you very much @henrygab for spending lots of time fixing this issue. This has been draining a lot more time than expected ( and I am got pulled to work on other things meanwhile). I am very much appreciated for your time and effort. Looking forward to see more of your PRs in the future :)

henrygab:
@hathach -- I am humbled by the many volunteers, without whose help (reporting, testing the many incomplete fixes, and final validation) this would not have been possible, including:
* @czuvich
* @matthewswilson
* @mattratto
* @jeremypoulter
* @jpconstantineau
* @shaoyuan1943

A special thanks also to @geky for helping me understand LFS!

mattratto:
Amazing! Great to watch this work being done!

hathach:
> @hathach -- I am humbled by the many volunteers, without whose help (reporting, testing the many incomplete fixes, and final validation) this would not have been possible, including:
> 
> * @czuvich
> * @matthewswilson
> * @mattratto
> * @jeremypoulter
> * @jpconstantineau
> * @shaoyuan1943
> 
> A special thanks also to @geky for helping me understand LFS!

Indeed, thank you everyone to help fixing this annoying bugs and be PATIENT with me :D

mattratto:
HI all, and specifically Henry, after months of no corruption issues, I have now seen 5 issues, exhibited with both Adafruit BSP 0.16.0 and with Adafruit BSP 0.20.5 (most recent release.) In each case, the error is the same LFS 494 Corrupted Directory blocks. It is possible that these devices had an older version of the BSP installed previously and thus the corruption may have occurred while using an older version of the Adafruit littlefs libraries. It is resolved by a format, and has not reoccurred in any of the above examples. However, I am wondering how I might insulate my code against corruption issues. 

My main problem is that when my code runs into a corruption issue, it locks up. Does anyone have any suggestions how I could prevent this? I would like to show an error message and potentially kick off a format when the code encounters file corruption, but I have no idea how to do this since the lock up occurs while executing a file open. 

thanks, 

Matt

mattratto:
In the interests of full disclosure, below is the code that is locking up when it encounters an error. If this is really a code rather than a library issue, I do apologize: 
InternalFS.begin(); //start file system
  File item(InternalFS);
  File dir("data/", FILE_O_READ, InternalFS);
    while( (item = dir.openNextFile(FILE_O_READ)) )
  {  
    String filename = item.name(); 
    if(filename.indexOf("data") == 0) {  
    uint32_t readlen;
    char buffer[999] = { 0 };
    readlen = item.read(buffer, sizeof(buffer));
    item.close();
    for (int e=0; e<readlen; e++) { //now parse and separate 
    Serial.print(buffer[e]); 
 }
 Serial.println("---"); 
 }
 }
dir.close();
}

mattratto:
<script src="https://gist.github.com/mattratto/af60c80131fb506d64221a539ba781d6.js"></script>

mattratto:
after further thought, am thinking this is a code issue rather than a library issue per se. For anyone who wanders by, I realized that the 'Internal_ListFiles' example does not lock up, but my code above does. The only difference is that the example code only grabs the file names, but I get the names but also try to read the file contents. So I am wondering if there is a way to check the file integrity prior to attempting to read - maybe in fact, the item.name read is returning an error code?

hathach:
@mattratto the file system should done the checking for us, once the FS is corrupted, it is  too hard to check that, since you need to understand how the files is stored on the flash and understand how these inode chain work etc... 

Are you still seeing corruption issue with latest library ?

mattratto:
@hathach I am getting file corruption error messages when running code with latest library. However, I do not know if they were created by the library, or occurred when running an older library, prior to update to 0.20.5 BSP. These show up when I run Internal_ListFiles example code and seem to be happening around the use of openNextFile. I went and looked at this function in Adafruit_LittleFS_file.cpp and found  on line 393: 

   (void)ret._open(filepath, mode); // return value is ignored ... caller is expected to check isOpened()

I couldn't find isOpened, but am now trying isOpen() on each item prior to proceeding: 

while( (item = dir.openNextFile(FILE_O_READ)) )
  {
    if (item.isOpen()){ 

I will test this and report back.

hathach:
@mattratto I see your problem now, since this PR is merged and closed. You should really re-test the library with newly format drive, if you still have issue with latest code then. Feel free to open an new issue, that make following it easier ( your could point a ref link to this if needed).

mattratto:
OK. Will do. Thanks @hahatch


Matt Ratto
Associate Professor
Bell University Labs Chair in Human-Computer Interaction
Faculty of Information
University of Toronto
@mattratto

PR #398, opened 2019-12-22
Display BT device name (if available)

bitbank2:
Added a function to parse the BT scan record for the advertised device name.


hathach:
Sorry for late response, I am a bit busy with other works. Why would you close the issue ?

bitbank2:
I closed it because I saw after writing this code that there is already that functionality in another part of the bluetooth library.

hathach:
OK, thanks for reply. Apparently we didn't do a good job of documenting :)

PR #407, opened 2020-01-06
WIP: Nano 33 BLE

rpavlik:
Per  https://github.com/adafruit/Adafruit_nRF52_Bootloader/issues/94 and https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/97 I'm interested in using the UF2 bootloader on my Nano 33 BLE. I got it working, though the stock Arduino board def uses bossac, which doesn't work with the adafruit NRF52 bootloader. To adjust for this, I added a new board in this board collection to use nrfutil dfu. The pin definitions are based on the original (though ported to this repo's technique), most everything else is related to or based on the Itsy or Feather nrf52832 definitions.

I have successfully run several Adafruit BLE demos, including the "Controller" example where adding these lines to the color packet handler lets you control the onboard RGB LED using the Bluefruit app:

```c++
    analogWrite(LEDR, 255 - red);
    analogWrite(LEDG, 255 - green);
    analogWrite(LEDB, 255 - blue);
```

No idea if there's appetite for merging this, but I figured I'd share and submit it at the very least.


ladyada:
thanks! this is cool, and we want to make sure we match folk's expectations.
would you be ok adding a readme.md on how people load the bootloader on and renaming this to "Arduino Nano 33 BLE (Unsupported)"?  that way we dont end up with confused folks who dont realize you need a jlink to load the bootloader before starting! :)

rpavlik:
Glad you like it!  I've pushed my latest commits, which include a README in the bootloader directory. I also tested flashing back to the original bootloader, which worked great with my Atmel-ICE - flashing the Adafruit bootloader also worked great, so those are the instructions I used. Pretty sure there are some open-source CMSIS-DAP programmer firmwares out there, so should be accessible for those with the technical skills.

I do have some photos of my (hacky) pogo pin jig that I haven't added yet, let me know if you think they'd be useful. (You can put a mechanical-only pogo pin 2 holes away from either side of the electrically-useful ones to go into the actual header holes for alignment, which is handy if you have a no-header version of the board. Today I didn't even clamp it when doing the bootloader since I had it down to a single command I could send with one hand.)

ladyada:
yeah add any pics you like :) when you're done we'll do a final review

henrygab:
@rpavlik Hi Ryan,

I was taking a look at the Nano 33 BLE board, but it appears that the [official docs disagree](https://github.com/arduino/ArduinoCore-nRF528x-mbedos/issues/46) on D7 and D8.  Specifically, the official [**board pinout diagram**](https://content.arduino.cc/assets/Pinout-NANOsense_v1.pdf) shows D7 is P0.09, and D8 is P0.10.  However, the official Arduino BSP v1.1.3 [variant.cpp](https://github.com/arduino/ArduinoCore-nRF528x-mbedos/blob/8b642cc27b7056029357ca1c63d20bd9435ea40c/variants/ARDUINO_NANO33BLE/variant.cpp#L12-L15) says they are 0.23 and P0.21, respectively:

https://github.com/arduino/ArduinoCore-nRF528x-mbedos/blob/8b642cc27b7056029357ca1c63d20bd9435ea40c/variants/ARDUINO_NANO33BLE/variant.cpp#L12-L15

Your commits appear to follow the BSP.  Can you verify that D7 and D8, when mapped to P0.23 and P0.21 respectively as in your commits, will work in blink sketch targeting D7 / D8?

If an LED doesn't blink on D7/D8 as-is, can you try again after adjusting `variant.cpp` to have D7 / D8 at  P0.09 / P0.10, respectively?

Thanks!

rpavlik:
Good catch! I'll have to solder in the headers and try that one out. Pretty sure it's not the first inconsistency I've seen in the docs.

henrygab:
@rpavlik -- Well, they [updated the picture](https://github.com/arduino/ArduinoCore-nRF528x-mbedos/issues/46) to match the variant files.  So, no need to change pins at this point.  :)

codeanpeace:
Any updates on this? Hoping to use an SDK built on top of this with the nano sense :)

ElreboCM:
This is very cool contribution. 
using the [Sparkfun Pogo Pin Adapter](https://www.sparkfun.com/products/11591) it’s quite easy to flash the nano 33 with a j-link. 

Does OTA also work with the bootloader?
That would be awesome!

rpavlik:
Oh that's a nice tool! Kind of like the one I built myself from perf board. I have never tried the ota stuff on an nrf chip so I don't know, sorry. I suspect that if it works on an Adafruit board, it probably will work with this conversion, but I am not sure. Recently I've been using mine not for BLE at all, but just as a fast processor with an attached imu.

ElreboCM:
I am wondering what is missing for the pull to be accepted?

PR #408, opened 2020-01-07
Lfs followup

hathach:
follow up to  #397, move lockFS/unlockFS to FileSystem class since it make more sense, and replace the native freeRTOS call with mutex (maybe helpful to port to other platform later on). Skip mutex for non-LFS function such as isDirectory(), isOpen(), name()

@henrygab for review

### DIFF
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
@@ -60,7 +60,7 @@ bool File::_open_file (char const *filepath, uint8_t mode)
     _file = (lfs_file_t*) rtos_malloc(sizeof(lfs_file_t));
     if (!_file) return false;
 
-    int rc = lfs_file_open(_fs->getFS(), _file, filepath, flags);
+    int rc = lfs_file_open(_fs->_getFS(), _file, filepath, flags);
 
     if ( rc )
     {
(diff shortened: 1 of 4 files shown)


libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -319,24 +331,17 @@ bool File::_isOpen(void)
 //            suddenly (unexpectedly?) have different values.
 char const* File::name (void)
 {
-  this->_LockFilesystem();
-  char const* ret = this->_name;
-  this->_UnlockFilesystem();
-  return ret;
+  return this->_name;
```
hathach: @henrygab in my defense to skip FS lock, we only guard again lfs_* call since lfs is not reentrant for now. Over time, lfs will follow fatfs to provide option/hook for locking. By that we can revert most of this implementation by using the lfs lock hook. 

The scenario when a same File object is used by user to write as file in one thread, but close and open as directory (or other filepath) in aother thread  etc .. should be guarded by application itself.

henrygab: OK.  It's a design choice, and I admit I was **_very_** conservative in the locking.  Your changes will allow two File objects to get a subset of metadata without locking:
* `isOpen()`
* `name()`
* `isDirectory()`

each of the above can be returned without accessing LFS, so this appears safe.

hathach:
Thanks for reviewing the PR :)

henrygab:
Of course!

If you would be OK with a private email conversation: pt or limor can provide you my personal email … I have general questions on the collaborator role, and want to respect the trust shown by the invite.

ladyada:
hiya - please email pt@ and myself if you have any questions, we're the email wranglers :)

Issue #352, opened 2019-10-12
[Bug] Why does PDM require manual IRQ triggering?


ladyada:
Ideally for the PDM library (adapted from Arduino's) we should have the IRQ auto-called when there's more data, but to make the example work, i have to manually call it.

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/PDM/examples/PDMSerialPlotter/PDMSerialPlotter.ino#L44

 Why is that - what is not set up to make the IRQ enabled? That is your mystery :D

henrygab:
@ladyada - It appears the library code for the PDM peripheral may have some significant bugs in how it handles buffers.  Please accept the below code review of the library's configuration and use of the PDM peripheral.  I would appreciate confirmation of my findings.

**_Unfortunately, I do not have an appropriate microphone, and thus cannot test the results._**  However, if you wish, I can generate the fixes that appear to be necessary, and submit a PR for your own testing.

<details><summary>Interaction between tasks, events, buffers</summary>

<P>
According to the product specifications for nRF52840, v1.1 @ line 236:

> As soon as the STARTED event is received, the firmware can write
> the next SAMPLE.PTR value (this register is double-buffered), to
> ensure continuous operation.
> When the buffer in RAM is filled with samples, an END event is
> triggered. The firmware can start processing the data in the buffer.
> Meanwhile, the PDM module starts acquiring data into the new buffer
> pointed to by SAMPLE.PTR, and sends a new STARTED event, so that the
> firmware can update SAMPLE.PTR to the next buffer address.

Putting this in plain English, the rules for the code (library) appear to be:
1. Before triggering `NRF_PDM_TASK_START`, set the `SAMPLE.PTR` value.
   (_The peripheral saves (double-buffers) the `SAMPLE.PTR` register
   for immediate use, and generates a `START` event._)
2. When receiving a `START` event, the `SAMPLE.PTR` should be set to
   the *next* buffer that will be used, after the double-buffered value
   that was set (and is the current value) is filled by the peripheral.
   (_The peripheral will ignore the value place in that register
   until **after** it fills the current buffer._)
3. When receiving an `END` event, that means the peripheral finished
   filling a buffer (which was double-buffered, and thus unlikely to
   be the same as currently in the `SAMPLE.PTR` register).

</details>

<details><summary>Library fails to set `SAMPLE.PTR` prior to triggering `NRF_PDM_TASK_START`</summary>

<P>

[...]

PR #409, opened 2020-01-08
fix pdm issue with #352

hathach:
- also add define for PIN_PDM_DIN, PIN_PDM_CLK, PIN_PDM_PWR
- add PDM setPins() for custom board

### DIFF
--- a/libraries/PDM/examples/PDMSerialPlotter/PDMSerialPlotter.ino
+++ b/libraries/PDM/examples/PDMSerialPlotter/PDMSerialPlotter.ino
@@ -10,10 +10,6 @@
 
 #include <PDM.h>
 
-// The default Circuit Playground Bluefruit pins 
-// data pin, clock pin, power pin (-1 if not used)
-PDMClass PDM(24, 25, -1);
-
 // buffer to read samples into, each sample is 16-bits
 short sampleBuffer[256];
 
@@ -41,8 +37,6 @@ void setup() {
 
 void loop() {
   // wait for samples to be read
-  PDM.IrqHandler();
-
   if (samplesRead) {
     // print samples to the serial monitor or plotter
     for (int i = 0; i < samplesRead; i++) {
@@ -63,3 +57,4 @@ void onPDMdata() {
   // 16-bit, 2 bytes per sample
   samplesRead = bytesAvailable / 2;
 }
+
--- a/libraries/PDM/src/PDM.cpp
+++ b/libraries/PDM/src/PDM.cpp
@@ -44,15 +44,28 @@ PDMClass::~PDMClass()
 {
 }
 
+void PDMClass::setPins(int dataPin, int clkPin, int pwrPin)
+{
+  _dinPin = dataPin;
+  _clkPin = clkPin;
+  _pwrPin = pwrPin;
+}
+
 int PDMClass::begin(int channels, long sampleRate)
 {
   _channels = channels;
 
-  /*
+  // Enable high frequency oscillator if not already enabled
+  if (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0) {
+    NRF_CLOCK->TASKS_HFCLKSTART    = 1;
+    while (NRF_CLOCK->EVENTS_HFCLKSTARTED == 0) { }
+  }
+
   // configure the sample rate and channels
   switch (sampleRate) {
     case 16000:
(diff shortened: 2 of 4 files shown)


libraries/PDM/src/PDM.cpp
```diff
@@ -203,13 +214,24 @@ void PDMClass::IrqHandler()
...
+
+#ifndef PIN_PDM_CLK
+#define PIN_PDM_CLK -1
+#endif
+
+#ifndef PIN_PDM_PWR
+#define PIN_PDM_PWR -1
+#endif
+
+PDMClass PDM(PIN_PDM_DIN, PIN_PDM_CLK, PIN_PDM_PWR);
```
hathach: PDM is declared within the class like Serial or Wire. It is consistent with current usage of PDM from Arduino nRF528x-mbedos. Making using 3rd party library easier. 

Note: it requires to remove this line from CPlay library https://github.com/adafruit/Adafruit_CircuitPlayground/blob/master/utility/Adafruit_CPlay_Mic.cpp#L30 . I will submit an PR for this after this once is merged.

libraries/PDM/src/PDM.h
```diff
@@ -33,6 +33,7 @@ class PDMClass
   PDMClass(int dataPin, int clkPin, int pwrPin);
   virtual ~PDMClass();
 
+  void setPins(int dataPin, int clkPin, int pwrPin);
```
hathach: add setPins(), will be useful for custom nrf52 board.

variants/circuitplayground_nrf52840/variant.h
```diff
@@ -134,6 +136,13 @@ static const uint8_t SCK  = PIN_SPI_SCK ;
 #define USB_MSC_BLOCK_SIZE    512
 #define USB_MSC_BLOCK_COUNT   ((2*1024*1024) / USB_MSC_BLOCK_SIZE)
 
+/*
+ * PDM Interfaces
+ */
+#define PIN_PDM_DIN         24
```
hathach: currently these pin are defined for CPB

libraries/PDM/examples/PDMSerialPlotter/PDMSerialPlotter.ino
```diff
@@ -10,10 +10,6 @@
 
 #include <PDM.h>
 
-// The default Circuit Playground Bluefruit pins 
-// data pin, clock pin, power pin (-1 if not used)
-PDMClass PDM(24, 25, -1);
```
hathach: removed, we could use PDM directly like we did with Serial

libraries/PDM/src/PDM.cpp
```diff
@@ -145,7 +158,7 @@ int PDMClass::available()
 
   size_t avail = _doubleBuffer.available();
 
-  //NVIC_EnableIRQ(PDM_IRQn);
+  NVIC_EnableIRQ(PDM_IRQn);
```
henrygab: By disabling the PDM in the NVIC, wouldn't this potentially cause the silent loss of samples of data?
Especially with softdevice enabled, and Bluetooth active, which can pre-empt the execution here and cause unexpectedly long delays while this is still disabled?

libraries/PDM/src/PDM.cpp
```diff
@@ -116,7 +128,6 @@ int PDMClass::begin(int channels, long sampleRate)
   nrf_pdm_event_clear(NRF_PDM_EVENT_STARTED);
   nrf_pdm_task_trigger(NRF_PDM_TASK_START);
```
henrygab: The register buffer pointer is still set to `nullptr`.  Can you help me understand why this might be intentional?

hathach:
> I think this may not result in a working PDM library.
> 
> It appears that buffers are provided (via the callback), before they are actually filled by the DMA engine.
> 
> I'd like to better understand how using the NVIC disable / enable for the public APIs will (or will not) risk losing samples.

@henrygab thank you very much for analyzing the PDM implementation, unfortunately I am not capable of answering your question since I just learned about PDM recently. I mostly ste..l hmm use the code from Arduino repo + code comparing. You should really file issue there to get answer from PDM writer.  
https://github.com/arduino/ArduinoCore-nRF528x-mbedos/blob/master/libraries/PDM/src/PDM.cpp

PR #410, opened 2020-01-09
Remove unused variable.

henrygab:
Simple fix for build warning

### DIFF
--- a/cores/nRF5/utility/SoftwareTimer.cpp
+++ b/cores/nRF5/utility/SoftwareTimer.cpp
@@ -44,7 +44,7 @@ bool SoftwareTimer::start(void)
   bool ret = true;
   if ( isInISR() )
   {
-    BaseType_t ret, xHigherPriorityTaskWoken = pdFALSE;
+    BaseType_t xHigherPriorityTaskWoken = pdFALSE;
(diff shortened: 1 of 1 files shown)


cores/nRF5/utility/SoftwareTimer.cpp
```diff
@@ -44,7 +44,7 @@ bool SoftwareTimer::start(void)
   bool ret = true;
   if ( isInISR() )
   {
-    BaseType_t ret, xHigherPriorityTaskWoken = pdFALSE;
+    BaseType_t xHigherPriorityTaskWoken = pdFALSE;
```
hathach: worse issue, this shadow the outter ret variable, thank you very much. This needs an eagle eye to catch :)

ladyada:
nice!

cores/nRF5/utility/SoftwareTimer.cpp
henrygab: No eagle eye required -- just build with all warnings enabled.  It's even an option exposed in the UI.  :)

However, it appears that the move away from the Travis builds lost the ability to find regressions related to warning levels.  Now, warnings are ignored by the GitHub action based builds.  (This may be how it slipped past?)

PR #423, opened 2020-01-16
Split debug menu and add Segger RTT as option for logging output

### LINKED ISSUE #416: [WIP] Split debug menu and add Segger RTT as option for logging output
I'm in process of preparing the solution for this issue.

**Is your feature request related to a problem? Please describe.**
Yes.  When attempting to use SysView, the large number of debug prints make it difficult to focus on the new areas.  The large quantity of debug prints also hide Heisenbugs, as due to significant timing changes, some bugs would not repro with SysView enabled.

Enabling Segger RTT as an option for higher-speed terminal output is also strongly desired, but the currently included RTT sources are both out-of-date and failed to include the necessary files.

**Describe the solution you'd like**
At least three parts:

* [ ] Boards.txt needs an update so that enabling SysView does not require a high debug level (or any debug level at all).  This will allow SysView to be used with minimal performance changes, and thus allow deeper analysis (including better performance analysis) of code using this platform.
* [ ] Update Segger RTT files and enable selection of serial port or RTT as default logging output.
* [ ] Either expose a "default logger" variable (that derives from Stream class), or override the global `Serial` to instead redirect to RTT when RTT is enabled in the UI.  

This second option of redirecting Serial to RTT would need additional validation and review of impacted locations in at least the BSP, such as ensuring firmware updates will still work over serial port if Serial ==> RTT.

henrygab:
Resolves #416 

At its core, this PR does three things, but does **_	not_** change the default build behavior (CFG_DEBUG=0).

1. Separates the concept of CFG_DEBUG (debug level in UI) from whether or not SysView is enabled.  As noted in #416, only enabling SysView with CFG_DEBUG set to 3+ can prevent SysView from being used to find Heisenbugs.  Now, SysView can be enabled even when CFG_DEBUG is set to zero.

2. Exposes a "default" logger that derives from the Stream class.  This default logger is used for all built-in board logging output, and is exposed for use by libraries if they choose to use it.  By default, this is still Serial, and thus has no appreciable effect when using default build configuration.

3. Add support for Segger RTT as the default data log, helping to track Heisenbugs that otherwise would be hidden when logging levels are increased.  Serial remains the default, and a new menu option is provided to select the use of RTT for board-level logging.

Tested on nRF52840 Feather Express hardware with various option combinations, and RTT tested both with and without JLink attached.

NOTE:
* `Serial.begin()` is no-op.  Why does main.cpp call it (around line 50)?
_[edit: not a no-op on nrf52832, which  uses hardware uart]_

_[edit Also tested building using `arduino-builder.exe` … made life so much easier!]_


henrygab:
Minor update -- Expose underlying class, rather than just a `Stream&`, for `Adalog_Default_Logger`.

hathach:
Thanks @henrygab your reason is all valid, however, I don't prefer to have 2 extra menu options just to deal with this. Especially there is only 5% of people will ever use it. This Arduino repo still primarily target beginner, most don't ever have a jlink to begin with. 

To sum up, I think we should still keep the menu as it is with a simple modification to automatically enabled **CFG_SYSVIEW** when **CFG_DEBUG=3**, else you could just modify the source to manually enable CFG_SYSVIEW alone while keeping debug level = 0. Same with logger with SYSVIEW, please add **CFG_SYSVIEW_LOG** somewhere in the source where you could just toggle 0/1 when needed.

I am really like your work, unfortunately, this PR will be only used by a very small users and could cause confusion to the vast majority.

henrygab:
Hi @hathach , and welcome back.  I hope your time off was excellent.

Based on your review, I am removing the menu options as requested.  I also updated `.gitignore` to ignore `platform.local.txt` and `boards.local.txt`, as they should never be checked in (at least to master).

In this way, if desiring to use the Segger RTT and/or Segger SysView, it is now easily done using either `platform.local.txt` (for always changed) or `boards.local.txt` (for menu-options in Arduino UI).

boards.txt
```diff
@@ -57,7 +57,7 @@ feather52832.menu.debug.l1.build.debug_flags=-DCFG_DEBUG=1
 feather52832.menu.debug.l2=Level 2 (Full Debug)
 feather52832.menu.debug.l2.build.debug_flags=-DCFG_DEBUG=2
 feather52832.menu.debug.l3=Level 3 (Segger SystemView)
-feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3
+feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3 -DCFG_SYSVIEW=0xFFFFFFFF
```
hathach: why CFG_SYSVIEW has to be max int, an simple 0,1 should be easier to follow.

boards.txt
```diff
@@ -338,7 +337,7 @@ pca10056.upload.protocol=nrfutil
 pca10056.upload.use_1200bps_touch=true
 pca10056.upload.wait_for_upload_port=true
 pca10056.upload.maximum_size=815104
-pca10056.upload.maximum_data_size=237568
+ -DCFG_SYSVIEW=0xFFFFFFFFpca10056.upload.maximum_data_size=237568
```
hathach: pca10056 is missing

cores/nRF5/freertos/config/FreeRTOSConfig.h
```diff
@@ -197,8 +197,13 @@ standard names - or at least those used in the unmodified vector table. */
 #define configUSE_DISABLE_TICK_AUTO_CORRECTION_DEBUG     0
 
 // Sysview require at least debug level 3
-#if CFG_DEBUG >= 3
-#include "sysview/SEGGER_SYSVIEW_FreeRTOS.h"
+#if CFG_SYSVIEW
+    #include "sysview/SEGGER_SYSVIEW_FreeRTOS.h"
+#endif
+
+// Segger RTT output requires at least debug level 20
```
hathach: comment is not relevant now.

cores/nRF5/sysview/SEGGER/SEGGER_RTT.c
```diff
@@ -52,15 +42,16 @@
 *                                                                    *
 **********************************************************************
 *                                                                    *
-*       SystemView version: V2.52d                                    *
+*       RTT version: 6.60d                                           *
```
hathach: This is not important but you seem to use the rtt file bundled with jlink. Since jlink version change very often, I much prefer the **Target Sources** download, since it is tested against the SystemView in specific version.

https://www.segger.com/downloads/systemview/

cores/nRF5/sysview/SEGGER/SEGGER_RTT.c
```diff
@@ -240,7 +235,7 @@ Additional information:
 **********************************************************************
 */
 
-static unsigned char _aTerminalId[16] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
+static unsigned char _aTerminalId[SEGGER_RTT_NUMBER_OF_TERMINALS] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
```
hathach: Please do not modify the original RTT file from SEGGER. We should leave it as it is to make it effort-free maintenance

cores/nRF5/sysview/SEGGER/SEGGER_RTT.h
```diff
@@ -83,6 +134,9 @@ Revision: $Rev: 12706 $
 **********************************************************************
 */
 
+#define SEGGER_RTT_NUMBER_OF_TERMINALS 16
```
hathach: Please do not modify the original RTT file from SEGGER. We should leave it as it is to make it effort-free maintenance

hathach:
> Hi @hathach , and welcome back. I hope your time off was excellent.

Thank you very much, it is indeed excellent, except for the Coronavirus ~ . ~

> Based on your review, I am removing the menu options as requested. I also updated `.gitignore` to ignore `platform.local.txt` and `boards.local.txt`, as they should never be checked in (at least to master).
> 
> In this way, if desiring to use the Segger RTT and/or Segger SysView, it is now easily done using either `platform.local.txt` (for always changed) or `boards.local.txt` (for menu-options in Arduino UI).

I don't really like this approach either, but it is doable. I would expect advanced user modify the source file directly without the help of IDE e.g

```
#if CFG_DEBUG > 3
#define CFG_SEGGER    1
#endif

#ifndef  CFG_SEGGER
#define CFG_SEGGER    0 // change to 1 to use with SYSVIEW without logging
#endif 
```

boards.txt
henrygab: Because longer term, I want to have SysView be a bitmask, so that you could only enable SysView events of specific types.  Already there are lost events with standard JLINK at debug level 3 for any but the simplest projects, and a JLINK EDU Mini is even slower.  So, reducing that output is a longer-term goal, so that it can be more useful by being targeted.  By doing this as 0xFFFFFFFF now, the default at debug level 3 need not be changed, allowing boards.txt to not require a change later.

I hope that explains my thoughts OK for this?

> Please revert the original RTT sources files, we will only use it as it is.

Why revert the RTT sources?  Without those, RTT does not work at all?  What changes of any value would remain?

boards.txt
henrygab: Fixed.

cores/nRF5/freertos/config/FreeRTOSConfig.h
henrygab: Fixed.  Good catch.

henrygab:
I will have to think about how to meet all the changes you would like.
Some things are easy.  Some less so.

boards.txt
hathach: > Because longer term, I want to have SysView be a bitmask, so that you could only enable SysView events of specific types. Already there are lost events with standard JLINK at debug level 3 for any but the simplest projects, and a JLINK EDU Mini is even slower. So, reducing that output is a longer-term goal, so that it can be more useful by being targeted. By doing this as 0xFFFFFFFF now, the default at debug level 3 need not be changed, allowing boards.txt to not require a change later.
> 
> I hope that explains my thoughts OK for this?

I see, that is too advanced, please move the `CFG_SYSVIEW` in to source file, where advanced user can decide which mask they can enable. By default, CFG_DEBUG=3 will also set it to 0xFFFFFFF as current.

> 
> > Please revert the original RTT sources files, we will only use it as it is.
> 
> What do you mean by "revert the original RTT source files"? Without those, RTT does not work at all, as not all RTT files were included before.

I mean we should only use Segger RTT sources file as it is when downloaded from segger. We should not modify it at all, that will make it difficult to upgrade later on.

cores/nRF5/sysview/SEGGER/SEGGER_RTT.h
```diff
@@ -150,6 +204,7 @@ unsigned     SEGGER_RTT_HasDataUp               (unsigned BufferIndex);
 void         SEGGER_RTT_Init                    (void);
 unsigned     SEGGER_RTT_Read                    (unsigned BufferIndex,       void* pBuffer, unsigned BufferSize);
 unsigned     SEGGER_RTT_ReadNoLock              (unsigned BufferIndex,       void* pData,   unsigned BufferSize);
+int          SEGGER_RTT_Peek                    (void);
```
hathach: we don't need to have RTT as Stream instance, we can change nanolib `_write()` hook to retarget to rtt.

henrygab:
Please see new PR #431.

It is a squashed merge from this branch, and thus will provide a cleaner history.

PR #431, opened 2020-02-01
[WIP] Update SysView / RTT

### LINKED ISSUE #416: [WIP] Split debug menu and add Segger RTT as option for logging output

henrygab:
Resolves #416 (now with @hathach's input on design)

Expose default logging output as Stream
reference, to enable use of Serial / RTT.

Enable RTT to emulate Serial class.
Add Peek() to SEGGER RTT.

At debug level 3, use explicit #define to
enable SYSVIEW (allows user to override
via boards.local.txt / platform.local.txt).

Extend RTT functions to enable `_write()`
to directly map to an RTT function.
(`SEGGER_RTT_TerminalOutBuffer()`)


henrygab:
@hathach, per your request, I have worked hard to avoid any change to the as-shipped SEGGER SysView files, and have updated them from the source you recommended.  I did need to deviate from the strict letter of this instruction only twice:

<details><summary>To disable the use of the assembly-based memcpy()</summary>

<P>

Required a modification to the file:
`cores\nRF5\sysview\Config\SEGGER_RTT_Conf.h`

to match the existing configuration, which set `RTT_USE_ASM` to zero:
`#define RTT_USE_ASM 0`
</details>

<details><summary> To extend RTT to emulate `Serial`, I left the file unmodified, but renamed it </summary>

<P>

This was so that the source file could be directly included in a different .c file, and not be compiled on its own.  Why was this necessary?  the file had static functions and variables that were necessary to use in the extension, but as you know, statically defined variables and functions are only accessible within the single compilation unit.  So, this hack allowed use in the extensions, without changing the source file itself.

That said, I think it is a clean solution given the constraints, as otherwise there would be no way to safely and temporarily write to a non-default terminal index … while the terminal index could be changed to a new value, because the current terminal index was not exposed, there was no other way to return to the prior value.   (without modifying the SEGGER code, which was not desired).

</details>

I hope this will be found more to your liking.

hathach:
> @hathach, per your request, I have worked hard to avoid any change to the as-shipped SEGGER SysView files, and have updated them from the source you recommended. I did need to deviate from the strict letter of this instruction only twice:

Thank you very much for putting more effort into this

> 
> To disable the use of the assembly-based memcpy()
> Required a modification to the file:
> `cores\nRF5\sysview\Config\SEGGER_RTT_Conf.h`
> 
> to match the existing configuration, which set `RTT_USE_ASM` to zero:
> `#define RTT_USE_ASM 0`

Why don't you just let it run as ASM ? It is probably faster than C code ?

>  To extend RTT to emulate `Serial`, I left the file unmodified, but renamed it
> This was so that the source file could be directly included in a different .c file, and not be compiled on its own. Why was this necessary? the file had static functions and variables that were necessary to use in the extension, but as you know, statically defined variables and functions are only accessible within the single compilation unit. So, this hack allowed use in the extensions, without changing the source file itself.
> 
> That said, I think it is a clean solution given the constraints, as otherwise there would be no way to safely and temporarily write to a non-default terminal index. Specifically, while it's possible to change the terminal index to a new value, there was no other way to read the existing terminal index, and therefore no way to return to the prior terminal index after printing (without modifying the SEGGER code, which was not desired).
> 
> I hope this will be found more to your liking.
> 
> _[edit: clarified language explaining on renamed file]_

I am really sorry, but I still don't like this HACK by including C file by another C file. Please drop the stream support for RTT if it can not be done. The Segger RTT and other external libraries should be as it is and upgrade to later version by the mean of copying files.

Sorry if am too picky, since this is only used by less than 5% of user, I don't even use it much since adding it a couple years ago out of curiosity.

boards.txt
```diff
@@ -57,7 +57,7 @@ feather52832.menu.debug.l1.build.debug_flags=-DCFG_DEBUG=1
 feather52832.menu.debug.l2=Level 2 (Full Debug)
 feather52832.menu.debug.l2.build.debug_flags=-DCFG_DEBUG=2
 feather52832.menu.debug.l3=Level 3 (Segger SystemView)
-feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3
+feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3 -DCFG_SYSVIEW=0xFFFFFFFF
```
hathach: Please remove all of `-DCFG_SYSVIEW=0xFFFFFFFF` , CFG_DEBUG level will by default enable it. This can done within an header/config while with a single ifdef rather than adding it to all of the board variant.

```
#if CFG_DEBUG >= 3
#define CFG_SYSVIEW   0xFFFFFFFF
#endif 
```

platform.txt
```diff
@@ -48,16 +48,19 @@ compiler.elf2hex.cmd=arm-none-eabi-objcopy
 compiler.ldflags=-mcpu={build.mcu} -mthumb {build.float_flags} -Wl,--cref -Wl,--check-sections -Wl,--gc-sections -Wl,--unresolved-symbols=report-all -Wl,--warn-common -Wl,--warn-section-align --specs=nano.specs --specs=nosys.specs
 compiler.size.cmd=arm-none-eabi-size
 
-# this can be overriden in boards.txt
+# these can be overriden in boards.txt
 build.float_flags=-mfloat-abi=hard -mfpu=fpv4-sp-d16 -u _printf_float
 build.debug_flags=-DCFG_DEBUG=0
+# CFG_LOGGER: 1 == Serial, 2 == RTT;  See enum in cores/nRF5/common_func.h
```
hathach: Let's move these CONFIGURE into the source/header file somewhere as well. I find it easier to modify the config .h/.c easier than editing the platform.txt

henrygab: Ok.

boards.txt
henrygab: done.

henrygab:
There are multiple items, so I will try to get close to addressing those not already resolved.  Please note the sections expand by clicking on the text next to the arrow.

> Why don't you just let it run as ASM ? It is probably faster than C code ?
<details><summary>build errors</summary>

<P>
Without that change, there were build errors.   The prior configuration has the configuration set to not use assembly.  The configuration files can (and are designed to be) changed when used.  This is totally normal.  Therefore, changing to not use ASM is meeting the goal of minimizing the change from the prior checked-in version.

</details>

> I am really sorry, but I still don't like this HACK by including C file by another C file.

<details><summary><B><U>Can you help me understand...?</U></B></summary>

<P>

I admit to some frustration here … so I've rewritten this many times to try to better understand the reasoning.

* The PR **_does_** enable an upgrade to a later version by copying the new files over the old files.  The process is already manual, and renaming a single file does not appreciably increase the difficulty of the process.

* By your own admission, you've never updated the Segger files … for multiple years.  Thus, even if every upgrade takes an extra ten minutes, this would not seem to be a major factor.

* If the concern is that requiring a file to be renamed would cause errors, this is simply untrue and strongly mitigated.  In fact, the PR makes the situation **_BETTER_** than it was previously, by ensuring a compilation error in `SEGGER_Extensions.h`, and right where that error would occur, gives actionable details on **_why_** it exists, **_when_** the file can be discarded, **_what_** to do when upgrading SEGGER SysView files, and even **_how_** to upgrade SEGGER files.

Based on the above, I am currently having trouble understanding the concerns on which this rejection is based.

The desire to not modify the files in any way also seems … unusually strong.

Can you help me understand if there is some other, not yet disclosed, reason for hesitation here?  Or if I've misunderstood some aspect of your concerns?

</details>

Before I deeply writeup the problems, design decisions, and constraints which drove this second revision, I'd like to know that we share a baseline understanding (common facts).  

<details><summary><B><U>Would you disagree with any of the following list?</U></B></summary>

<P>

**A.** Some bugs will fail to reproduce when the timing of execution changes significantly.  These types of difficult-to-find bugs have been called ['Heisenbugs'](https://en.wikipedia.org/wiki/Heisenbug).

**B.** The Feather nRF52840 is the only Arduino-targeted feather that includes a ten-pin SWD header as-shipped.

**C.** Some projects and 3rd party libraries do use `printf()` and related functions.  For those uses of `printf()`, the override of `_write(int file, char *ptr, int len)` function currently redirects to Serial.  The PR can (and already did) redirect to RTT as appropriate.

**D.** In the Arduino ecosystem, 3rd party library **examples** nearly universally use `Serial` as the debug and/or console output (for better or worse).

**E.** In the Arduino ecosystem, nearly all example sketches and instructional material use `Serial` for debug and/or console output.

**F.** The `Serial` object has a de-facto API that is so entrenched that, even when a new `Serial` object is defined that is not a hardware UART (e.g., the USB CDC Serial), the de-facto API is fully implemented.  This is true, even where the functions don't make sense, such as setting setting baud rate on USB CDC implementation.

**G.** The following function is required by the `Print` class:
  ```Arduino
      virtual size_t write(const uint8_t *buffer, size_t size);
  ```
  Therefore, any object that derives from `Print`, including `Stream` and `Serial`, must provides these functions.  It is significant that the buffer provided to this function is not required to be null-terminated.

**H.** The Feather boards are targeted at Arduino (and CircuitPython) users.  Of those users, <5% use the SWD header.

**I.** Debug output is increased at higher debug levels.

</details>

As you can see, the baseline is not anything controversial.  :)

hathach:
> There are multiple items, so I will try to get close to addressing those not already resolved. Please note the sections expand by clicking on the text next to the arrow.
> 
> > Why don't you just let it run as ASM ? It is probably faster than C code ?
> 
> build errors
> Without that change, there were build errors. The prior configuration has the configuration set to not use assembly. The configuration files can (and are designed to be) changed when used. This is totally normal. Therefore, changing to not use ASM is meeting the goal of minimizing the change from the prior checked-in version.

Thanks for explanation, I am fine with either C or ASM.

> > I am really sorry, but I still don't like this HACK by including C file by another C file.
> 
> Can you help me understand...?
> I admit to some frustration here … so I've rewritten this many times to try to better understand the reasoning.
> 
> * The PR **_does_** enable an upgrade to a later version by copying the new files over the old files.  The process is already manual, and renaming a single file does not appreciably increase the difficulty of the process.
> * By your own admission, you've never updated the Segger files … for multiple years.  Thus, even if every upgrade takes an extra ten minutes, this would not seem to be a major factor.
> * If the concern is that requiring a file to be renamed would cause errors, this is simply untrue and strongly mitigated.  In fact, the PR makes the situation **_BETTER_** than it was previously, by ensuring a compilation error in `SEGGER_Extensions.h`, and right where that error would occur, gives actionable details on **_why_** it exists, **_when_** the file can be discarded, **_what_** to do when upgrading SEGGER SysView files, and even **_how_** to upgrade SEGGER files.
> 
> Based on the above, I am currently having trouble understanding the concerns on which this rejection is based.
> 
> The desire to not modify the files in any way also seems … unusually strong.
> 
> Can you help me understand if there is some other, not yet disclosed, reason for hesitation here? Or if I've misunderstood some aspect of your concerns?
> 
> Before I deeply writeup the problems, design decisions, and constraints which drove this second revision, I'd like to know that we share a baseline understanding (common facts).
> 
> Would you disagree with any of the following list?
> As you can see, the baseline is not anything controversial. :)

Yeah, I added sysview out of curiosity and requirement to troubleshoot initial issues since freeRTOS make things harder to debug. Afterwards, I haven't used it much for years and have no reason to upgrade it. It is not the betterment to resist your PR, it is my rule of thumb when using external modules. Also I don't see making RTT as Stream is absolutely needed, we can just retarget ADA_LOG, the Serial output is user message, and have no impact when debugging.   

I understand you put lots of efforts and time into this PR. Time is valuable and I appreciate everyone effort when spending their time for this repo. However, I still think we should keep RTT module as it is and reduce the requirement to make it a Stream class to do so.

henrygab:
Marking as work-in-progress, per @hathach  feedback, as it will require another go-around.

hathach:
> Marking as work-in-progress, per @hathach feedback, as it will require another go-around.

Thanks Henry, I will try to help later on, when doing with nrf crypto and LESC work later on.

PR #435, opened 2020-02-05
Particle Xenon Support: Updates and fixes

outlandnish:
Adds boards.txt info for flashing Xenon from Arduino IDE
Fixes pin mappings for digital pins
Can now control RGB LED and Primary LED with` HardwarePWM` and `analogWrite()`


variants/particle_xenon/variant.cpp
```diff
@@ -27,33 +27,43 @@
...
+  _PINNUM(0, 11),   // P0.11 (MODE)
+  _PINNUM(0, 18),   // P0.18 (RESET)
 };
 
-static void switch_antenna(bool useExternal) {
+void initVariant()
+{
+  switch_antenna(false);
+
+  led_pwm_init(LED_PRIMARY_IDX, LED_PRIMARY_PIN);
```
DirtyEngineer: Error during compilation for missing function "led_pwm_init"

variants/particle_xenon/variant.cpp
```diff
@@ -27,33 +27,43 @@
...
+  _PINNUM(0, 29),   // P0.29 (PWM2) (ADC3)
+  _PINNUM(0, 28),   // P0.28 (PWM2) (ADC2)
+  _PINNUM(0, 4),    // P0.04 (PWM2) (ADC1)
+  _PINNUM(0, 3),    // P0.03 (PWM2) (ADC0)
+  _PINNUM(0, 11),   // P0.11 (MODE)
+  _PINNUM(0, 18),   // P0.18 (RESET)
 };
 
-static void switch_antenna(bool useExternal) {
+void initVariant()
```
DirtyEngineer: Duplicate function name below.

variants/particle_xenon/variant.h
```diff
@@ -182,8 +182,24 @@ static const uint8_t SCK1  = PIN_SPI1_SCK;
 
 // On-board QSPI Flash
 #define EXTERNAL_FLASH_DEVICES   GD25Q16C
+
+#define USB_MSC_BLOCK_SIZE    512
+#define USB_MSC_BLOCK_COUNT   ((2*1024*1024) / USB_MSC_BLOCK_SIZE)
+
 #define EXTERNAL_FLASH_USE_QSPI
 
+static void switch_antenna(bool useExternal);
```
hathach: static usage should be declared here, please remove this prototype

variants/particle_xenon/variant.h
```diff
@@ -182,8 +182,24 @@ static const uint8_t SCK1  = PIN_SPI1_SCK;
...
+
+// led pwm
```
hathach: these functions header and implementation shouldn't be included by variant. These should be called by user sketches using PWM module or analogWrite(). I did remove these recently.

README.md
```diff
@@ -109,6 +110,20 @@ $ nrfjprog --program feather_nrf52832_bootloader.hex -f nrf52
 $ nrfjprog --reset -f nrf52
 ```
 
+## Particle Xenon Specific
+```c++
+// Swap between the internal antenna and external u.FL antenna
+switchAntenna(bool useExternal);
```
hathach: Please remove these all xenon-specific functions, user should be able to do that in user sketch

variants/particle_xenon/variant.h
outlandnish: Is there a place you'd recommend documenting that for a user? It's not obvious how to to switch antennas otherwise.

variants/particle_xenon/variant.cpp
outlandnish: Yeah, I just noticed that. Removing it in the next commit

variants/particle_xenon/variant.cpp
outlandnish: I'm going to see if there's a way to incorporate this with analogWrite somehow

variants/particle_xenon/variant.h
outlandnish: @hathach analogWrite doesn't appear to do anything. Is it confirmed to work on other nRF52840 variants?

hathach: switching antennas should be on Particle documentation. You could define the PIN_ANTENNA_N and example sketch to do that. variant should be about PIN definition and its on-board usage.

analogWrite work with 52840, check out this example here https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/Bluefruit52Lib/examples/Hardware/Fading/Fading.ino .

outlandnish: That code you linked to does not work on the Particle Xenon. The original reason I added those led helper functions was to solve this very issue. For some reason, the existing analogWrite functions don't work on the Xenon.

hathach: please find out why it doesn't work. Either way, these helpers still shouldn't be included in variant.

outlandnish: Ok, I found out what was going on. It looks like the analogWrite and hardware pwm functions were using the `g_ADigitalPinMap` to figure out what pin to use (as opposed to the direct pin mappings in variant.h). I added all of the pins to `g_ADigitalPinMap` and now analogWrite works. I've removed the extraneous code from variant.cpp

outlandnish:
As mentioned in the changes, the primary and RGB leds now work with analogWrite and HardwarePWM

variants/particle_xenon/variant.cpp
```diff
@@ -27,33 +27,43 @@
...
-  35,   // P1.03 (UART2_RTS)
+  _PINNUM(0, 26),   // P0.26 (SDA)
+  _PINNUM(0, 27),   // P0.27 (SCL)
+  _PINNUM(1, 1),    // P1.01 (SDA1) (UART1_RTS) (SPI1_SCK) (PWM3)
+  _PINNUM(1, 2),    // P1.02 (SCL1) (UART1_CTS) (SPI1_MOSI) (PWM3)
+  _PINNUM(1, 8),    // P1.08 (UART2_TX) (SPI1_MISO) (PWM1)
+  _PINNUM(1, 10),   // P1.10 (UART2_RX) (PWM1)
+  _PINNUM(1, 26),   // P1.11 (UART2_CTS) (PWM1)
+  _PINNUM(1, 26),   // P1.12 (PWM0)
+  _PINNUM(1, 26),   // P1.03 (UART2_RTS)
```
jaswope: These are incorrectly defined. Should be 
```suggestion
  _PINNUM(1, 11),   // P1.11 (UART2_CTS) (PWM1)
  _PINNUM(1, 12),   // P1.12 (PWM0)
  _PINNUM(1, 3),   // P1.03 (UART2_RTS)
```
After making this change I was able to successfully get the analog_write sketch working on the blue on-board LED when specifying it as D7. It worked without the changes when using the LED_BUILTIN constant.

jaswope: These are incorrectly defined. Should be 
```suggestion
  _PINNUM(1, 11),   // P1.11 (UART2_CTS) (PWM1)
  _PINNUM(1, 12),   // P1.12 (PWM0)
  _PINNUM(1, 3),   // P1.03 (UART2_RTS)
```
After making this change I was able to successfully get the analog_write sketch working on the blue on-board LED when specifying it as D7. It worked without the changes when using the LED_BUILTIN constant.

outlandnish: Yep, I was able to address these changes (amongst other fixes) in the latest commit I pushed up. Thanks!

jaswope: They appear to still be misdefined in the latest commit. This would be D6, D7, and D8, and they are currently all defined as P1.26, instead of P1.11, P1.12, and P1.03 respectively.

outlandnish: I suppose it helps if I committed those changes 😫. Pushed that up in the latest commit.

variants/particle_xenon/variant.h
```diff
@@ -69,98 +70,91 @@ extern "C"
...
+static const uint8_t A5 = PIN_A5;
 #define ADC_RESOLUTION    14
 
 /*
  * Serial interfaces
  */
-#define PIN_SERIAL1_RX         PIN_D9
-#define PIN_SERIAL1_TX         PIN_D10
+#define PIN_SERIAL_RX         PIN_D9
+#define PIN_SERIAL_TX         PIN_D10
```
jaswope: I can't get this to compile unless I change this back to Serial1. It appears the Uart code expects Serial1 unless we're on a NRF52832: https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/Uart.cpp#L244-L248

outlandnish: Interesting, I get compiler errors when I change it back to Serial1. I'll play around with that

jaswope: I'm just directly checking out your branch and loading it locally via the Arduino IDE/VS Code. (pretty much following the instructions here: https://github.com/adafruit/Adafruit_nRF52_Arduino#optional-core-development-adafruit-nrf52-bsp-via-git)

jaswope: And looking at the Uart.cpp file, it definitely *should* be Serial1. I'm not sure how the build is succeeding without it.

outlandnish: It looks like PlatformIO is using an earlier tagged release of this library which appears to be why it works for me. I'll change it back to Serial1 and go through the steps to get PlatformIO to update their version

outlandnish:
Added the Adafruit bootloader for the Particle Xenon to the bootloader folder to facilitate builds

variants/particle_xenon/variant.h
hathach: should be Serial1 to be used with 52840 for actual UART, Serial on 840 is USB CDC

variants/particle_xenon/variant.h
```diff
@@ -20,6 +20,7 @@
 #define _VARIANT_PARTICLE_XENON_
 
 #define _PINNUM(port, pin)    ((port)*32 + (pin))
+#define NRF52
```
hathach: please remove this, you should either use NRF52_SERIES or NRF5232_XXAA or NRF52840_XXAA, which is std Nordic macros to detect nrf variants.
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/platform.txt#L60

variants/particle_xenon/variant.h
```diff
@@ -154,13 +148,13 @@ static const uint8_t A5  = PIN_A5 ;
...
 
-#define PIN_SPI_SS           PIN_A0
-#define PIN_SPI_MISO         _PINNUM(1, 14)
-#define PIN_SPI_MOSI         _PINNUM(1, 13)
-#define PIN_SPI_SCK          _PINNUM(1, 15)
-#define PIN_SPI1_MISO        _PINNUM(1, 8)
-#define PIN_SPI1_MOSI        _PINNUM(1, 2)
-#define PIN_SPI1_SCK         _PINNUM(1, 1)
+#define PIN_SPI_SS                  PIN_A0
+#define PIN_SPI_MISO                _PINNUM(1, 14)
```
hathach: all PIN_ in variant is mapped through `g_ADigitalPinMap`, you shouldn't use _PINNUM() for these peripherals, please double check it, it should use the Dx, or simply and index to the mapping array. 
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/SPI/SPI.cpp#L67

Check out feather nrf52840 for example, 52832 is only version we do direct mapping https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/variants/feather_nrf52840_express/variant.h

variants/particle_xenon/variant.h
```diff
@@ -175,13 +169,17 @@ static const uint8_t SCK1  = PIN_SPI1_SCK;
  */
 #define WIRE_INTERFACES_COUNT 2
 
-#define PIN_WIRE_SDA         _PINNUM(0, 26)
-#define PIN_WIRE_SCL         _PINNUM(0, 27)
-#define PIN_WIRE1_SDA        _PINNUM(1, 1)
-#define PIN_WIRE1_SCL        _PINNUM(1, 2)
+#define PIN_WIRE_SDA               _PINNUM(0, 26)
```
hathach: same as SPI

variants/particle_xenon/variant.h
```diff
@@ -69,78 +70,71 @@ extern "C"
...
+
+static const uint8_t A0 = PIN_A0;
+static const uint8_t A1 = PIN_A1;
+static const uint8_t A2 = PIN_A2;
+static const uint8_t A3 = PIN_A3;
+static const uint8_t A4 = PIN_A4;
+static const uint8_t A5 = PIN_A5;
 #define ADC_RESOLUTION    14
 
 /*
```
hathach: minor thing, you don't have to implement this, xenon has Serial2 mapping, we could put it as 
PIN_SERIAL2_TX/RX here, Serial2 is supported with latest release.

variants/particle_xenon/variant.h
outlandnish: Sounds good, will do! I had put that in for testing and forgot to remove it.

variants/particle_xenon/variant.h
outlandnish: Fixed!

variants/particle_xenon/variant.h
outlandnish: And added :)

variants/particle_xenon/variant.h
outlandnish: Fixed!

outlandnish:
@hathach @jaswope @DirtyEngineer thanks for pointing me in the right direction! Up until now I had changed just enough to get it to work for me but this helps a lot having it work properly.

jaswope:
Thanks for getting this off the ground. With Particle discontinuing support for the board, this is much needed. Please check out https://github.com/outlandnish/Adafruit_nRF52_Arduino/pull/1 for a few more changes that I thought would be appropriate for this merge.

outlandnish:
@jaswope yeah it's a bit sad that Particle is ending support (it's such a cheap and solid variant!) but it's really awesome they made as much of the documentation public as they did. That's what's really enabled any of this.

.github/workflows/githubci.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arduino-platform: ['feather52832', 'feather52840', 'feather52840sense', 'cplaynrf52840', 'itsybitsy52840', 'cluenrf52840' ]
+        arduino-platform: ['feather52832', 'feather52840', 'feather52840sense', 'cplaynrf52840', 'itsybitsy52840', 'cluenrf52840', 'particle_xenon' ]
```
hathach: Please remove this, ci for Adafruit boards for jow

outlandnish: Fair enough, done

jaswope: Sorry about that, my fault. It was a convenience for me and I meant to ask if it was appropriate or not.

DirtyEngineer:
Just a question,  Line 45 of variant.h
#define NUM_ANALOG_OUTPUTS   (6)

Should this not be "0" like all of the other nRF52840 boards?

outlandnish:
>  
> Just a question, Line 45 of variant.h
> #define NUM_ANALOG_OUTPUTS (6)
> 
> Should this not be "0" like all of the other nRF52840 boards?

@DirtyEngineer Are PWM outputs not considered analog outputs? If not, I can set that to 0.

DirtyEngineer:
@outlandnish It's my understanding that an true analog output would be from a DAC which to my knowledge the nRF52840 doesn't have. I think that the analog ouput field is legacy from the SAMD port which those chips do have a DAC. Anyone else have input on this?

jaswope:
I've been cribbing off of the Feather nRF52840 Express variant for much of this, and it has 0 for the number of analog output pins as well. My bet is this is the same.

jaswope:
In case it comes up: I defined the external flash device as a MX25L3233F (4MB flash), and I believe that to be correct. There's some differing information about which chip it uses when researching it, with some places saying it only has the 2MB MX25L1606 instead.

The reason I think the 4MB chip is correct is due to [this blog post](https://blog.particle.io/2018/10/23/particle-mesh-is-shipping-everything-you-need-to-know/) and the [datasheets included in the Xenon repo](https://github.com/particle-iot/xenon/tree/master/datasheets).

outlandnish:
Sounds good, updated

variants/particle_xenon/variant.h
```diff
@@ -175,13 +191,17 @@ static const uint8_t SCK1  = PIN_SPI1_SCK;
...
+#define PIN_WIRE_SCL               1
+#define PIN_WIRE1_SDA              2
+#define PIN_WIRE1_SCL              3
 
 // On-board QSPI Flash
-#define EXTERNAL_FLASH_DEVICES   GD25Q16C
+#define EXTERNAL_FLASH_DEVICES   MX25L3233F
+
+#define USB_MSC_BLOCK_SIZE    512
+#define USB_MSC_BLOCK_COUNT   ((2*1024*1024) / USB_MSC_BLOCK_SIZE)
```
jaswope: It looks like these definitions were recently removed in 3d49611c004e0569ce2b10c7e312803bffc69183, so we may not need them here.

.github/workflows/githubci.yml
hathach: @jaswope you have nothing to sorry about, we didn’t specify it explicitly anyway. Thanks for changes

variants/particle_xenon/variant.h
hathach: @jaswope yeah right, I will remove that later. Flash device is dynamically detected now.

variants/particle_xenon/variant.cpp
```diff
@@ -27,44 +27,56 @@
...
+  _PINNUM(0, 30),   // P0.30 (PWM3) (ADC4)
+  _PINNUM(0, 29),   // P0.29 (PWM2) (ADC3)
+  _PINNUM(0, 28),   // P0.28 (PWM2) (ADC2)
+  _PINNUM(0, 4),    // P0.04 (PWM2) (ADC1)
+  _PINNUM(0, 3),    // P0.03 (PWM2) (ADC0)
+  _PINNUM(0, 11),   // P0.11 (MODE)
+  _PINNUM(0, 18),   // P0.18 (RESET)
+
+  // LEDS
+  _PINNUM(1, 12),   // P1.12 (PRIMARY_LED)
```
henrygab: This doesn't seem right... 
* `D7  == _PINNUM(1,12) == PWM0`
* `D22 == _PINNUM(1,12) == PRIMARY_LED`

I don't think two Arduino pins should ever be the same underlying physical pin, as it can make for some really confusing bugs …

hathach: Right, these shouldn't be the same. @outlandnish @jaswope Since the board has pinout for D7, we should remove the D22, please check to see if @henrygab comment is correct whenever you have time.

outlandnish: Addressed in PR https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/440

PR #440, opened 2020-02-17
Xenon Support: Removed LED pin 22

outlandnish:
Removed duplicate pin mapping for the PRIMARY LED. Moved up all other pin mappings to account for the removed one.


hathach:
@henrygab no problems at all. We are all busy :). Thanks for your review

PR #452, opened 2020-03-11
Add more sensor services

hathach:
It turns out the PR add more services than just sound, the list is:
- Sound
- Color
- Proximity
- Gesture

Also other minor fix
- force while(!Serial) when debug level is enabled.
- fix feather sense D13 LED pin map


ladyada:
@hathach oops i think we're missing a dependency?

hathach:
> @hathach oops i think we're missing a dependency?

@ladyada  Yeah, I am fixing ci, please hold a bit, will ping you once it is ready to merge :)

hathach:
@ladyada it is fixed and ready to merge :ok_hand:

PR #453, opened 2020-03-11
[WIP] Groundwork for debug output improvements

henrygab:
This should be an easy PR to review.

### Important bits:

1. use uppercase PRINTF throughout the depot, rather than lowercase PRINTF.  This enables later (higher-performance) hook at the PRINTF level, rather than per-character _write() level.
2. For output previously going out through `Serial.print` and `Serial.println`, convert the code to use a reference to a `Print` class object (which `Serial` derives from).

### Useful info
I have compiled from commit 569b9b2 (was Adafruit master a few days ago) and from this PR's head (commit 1dcefd7), both at debug level 0 and 3.  I then analyzed the binaries using Ghidra and BinDiff, which validated that this PR does not change the compiled code.

### Note
This work is **_NOT_** limited to use only with Segger RTT, although that is definitely the first target.  Rather the longer target is to enable simple redirection of debug output,  to any object that supports the `Print` class.

This PR is a pre-requisite to enabling alternate debug output, such as Segger RTT, but also could be any other object that derives from the `Print` class (e.g., alternative serial port, alternative debugging output, TFT, LCD, etc.)

### DIFF
--- a/cores/nRF5/common_func.h
+++ b/cores/nRF5/common_func.h
@@ -142,7 +142,7 @@ const char* dbg_err_str(int32_t err_id); // TODO move to other place
 
 #define PRINT_LOCATION()      PRINTF("%s: %d:\n", __PRETTY_FUNCTION__, __LINE__)
 #define PRINT_MESS(x)         PRINTF("%s: %d: %s \n"   , __FUNCTION__, __LINE__, (char*)(x))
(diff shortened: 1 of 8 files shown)


henrygab:
So glad that CI is setup ... I believe this was caused by a setting in my platform.local.txt. I will be checking and updating the title as soon as I confirm just providing basic default values in platform.txt builds correctly.

henrygab:
Will rebase and ensure CI builds succeed.  I through CI was enabled in my prior master branch, but I was mistaken ... seems CI was temporarily disabled for a while.  Will open new PR from the new branch when done.

PR #454, opened 2020-03-12
Final prerequisites to improve debugger output

henrygab:
@hathach -- A simple PR for your consideration.

This PR provides the foundation to easily redirect debug output to an alternate output.  It standardizes use of the existing PRINTF for BSP output, while leaving example sketches unchanged.

Once merged, it will allow next PR's to focus on enabling alternative debug output devices (e.g., TFT, LCD, RTT, ...) -- so long as they derive from the `Print` class.

Summary of key changes:

1. Uses PRINTF instead of (stdio) printf in all core code (excludes examples).  There is effectively zero change to compiled code (validated w/BinDiff) through this change.

2. Modify existing hard-coded use of `Serial.print`, `Serial.println`, `Serial.printf` functions.  These now have a line similar to `Print& logger = Serial`, and then use `logger` instead of `Serial`.  There is no change to the final binary generated with this change.

3. Enable choice in how to override output mode (e.g., boards.local.txt, platform.local.txt, etc.).  There is zero change to generated binaries caused by this change.

4. The first commit fixes a debug level check from `==` to `>=`, but even this is a compile-time evaluation, and because the debug level cannot (in current sources) be greater than 3, this is create zero change to binaries generated with this change.

Thus, I believe this PR to be very low-impact.


platform.txt
```diff
@@ -51,13 +51,16 @@ compiler.size.cmd=arm-none-eabi-size
 # this can be overriden in boards.txt
 build.float_flags=-mfloat-abi=hard -mfpu=fpv4-sp-d16 -u _printf_float
 build.debug_flags=-DCFG_DEBUG=0
+build.logger_flags=-DCFG_LOGGER=1
```
hathach: @henrygab can we skip to include CFG_LOGGER and CFG_SYSVIEW in the `platforms.txt`. Let's just enable them based on teh CFG_DEBUG value. If you want to enable it in other scenario, just modify the source directly. Would that make sense ??

henrygab: Hi @hathach,
Thank you for the review.

Two questions, each with their own answer.  Taking each in turn...

<details><summary>As to CFG_LOGGER...</summary><P>

**Summary good news:**  As requested, I have no plans to update the `boards.txt` file to override these in the BSP.  As you indicate, the value can be updated by advanced users by editing the source code of the BSP.

At the same time, it is desirable to ensure Arduino-standard extensibility points remain open, especially when they are particularly useful in classroom settings.   Let me explain...

<summary>Extensibility and why the default value was added<details><P>

As you know, Arduino allows a overriding settings via a `boards.local.txt` file.  This is **_an extremely useful setting_** for a teacher / IT staff to enable custom menu options, although I understand that you yourself would never use this file.  This functionality is an Arduino standard, and part of the flexibility that the Arduino ecosystem provides -- IT staff / teachers can provide a customized `boards.local.txt` to allow students to:
1. Use the Board Manager in the Arduino UI without restriction, and
2. Simultaneously add menu options to the Arduino UI that customize an existing board

Here, the long-term goal is to enable choice of alternative (e.g., higher-speed, lower-overhead) debug logging via this extensibility mechanism ... Critically, **_without editing the BSP files_**.  After all, having a classroom use git (instead of the Board Manager) is not a desirable outcome ... it just adds too many layers of complexity when the goal is to teach coding.  At the same time, they may need both Serial ports for their project, and thus need to redirect the debug output to an LCD, to a file, etc.

While the entire line could **_technically_** be duplicated in an IT-provided `platform.local.txt`, the maintenance complexity is directly proportional to the length of the lines being added.  The `build.flags.nrf` setting line is one of the longest, and thus it is preferable to lay the groundwork to provide an alternative.  (_Yes, I know you don't use this feature._)

**I initially simply attempted to add the variables `{build.logger_flags} {build.sysview_flags}` in , _without providing a default value_.**  Really.  It's in the commit history.

Unfortunately, without a default value, compilation fails.  Therefore, a default value was necessary to avoid encumbering the Arduino `.local.txt` feature.

Because having a default value does not cause any compilation or maintenance problem, and because I am supporting your request not to add these options to the BSP's `boards.txt` list of options, this has no effect.  Thus, I added this commit, not because I forgot or ignored your earlier request to use the sources files (rather than the BSP's `boards.txt` file), but as a minimum necessary compromise to achieve your request while ensuring a smooth path forward for the goal of enabling alternate debug outputs.

BUT DO NOT FEAR...
</details>

Finally, I want to assure you that, although I had to add the default value into `platforms.txt` to allow this to build, I have every intention of continuing to support your request to **_NOT_** change the BSP's `boards.txt` -- I will not add the extra debug menu items there unless you request it, which is in line with your earlier feedback.

</details>
<details><summary>As to CFG_SYSVIEW...</summary><P>

I'm glad you understand that enabling of SYSVIEW may be desirable, without also enabling such verbose logging that a debug level of 3+ causes.  As you can see, this PR **_does not change_** the existing behavior:

* CFG_SYSVIEW is disabled by default
* `boards.txt` options enable SYSVIEW when debug level is set to 3+

As noted above wrt CFG_LOGGER, this PR lays the groundwork to allow this advanced functionality to be enabled, without requiring the user to edit BSP files.  This is important in classrooms and maker spaces, and even for online tutuorials.  (again, see more comprehensive explanation in prior section on the Arduino extensibility points).

Therefore, please accept this small addition, as it is necessary to simplify use of the standard Arduino extensibility points.

</details>

The Adafruit Feather nRF52840 Express is the only Feather with the built-in 10-pin debug header, and so is uniquely positioned for advanced classes, and provides a wonderful bridge when advancing users from Arduino to more advanced environments { platform.io, Segger Embedded Studio }.  I like this board.

The PR has **_zero functional change_**, has **_zero GUI changes_** (no menu options), and simultaneously makes forward progress to enable Arduino extensibility points for advanced class use.

Furthermore, I am actively ensuring your prior (and current) concerns are being addressed.  I have no plan for the next PRs to add any menu items, or to otherwise change the default logger of the BSP.

Please reconsider!

henrygab:
> everything is good, I only like to have only `CFG_DEBUG` in the platforms.txt CFG_LOGGER and CFG_SYSVIEW should be part of the source code if possible.

Thank you for the review.  I agree, and initially did not include the default value in `platform.txt`.  See above response for details on why this ended up included.

Yes, I understand this is not preferred, but I hope you can reconsider a compromise here, as this causes no GUI change, and no change to compiled code.

Please?  😃

platform.txt
hathach: @henrygab thanks for detail explanation, I am not aware of the use case of board/platforms.local.txt in the class/workshop as you mentioned. That is indeed a good scenario. Can I have my last attempt to ask you to bundle the `build.logger_flags + build.sysview_flags` into 1 flag, maybe **build.rtt_flags**, **build.extra_debug_flags**, or anything that makes sense to you e.g

```
build.extra_debug_flags=-DCFG_DEBUG=0 -DCFG_SYSVIEW=0
```

hathach:
> > everything is good, I only like to have only `CFG_DEBUG` in the platforms.txt CFG_LOGGER and CFG_SYSVIEW should be part of the source code if possible.
> 
> Thank you for the review. I agree, and initially did not include the default value in `platform.txt`. See above response for details on why this ended up included.
> 
> Yes, I understand this is not preferred, but I hope you can reconsider a compromise here, as this causes no GUI change, and no change to compiled code.
> 
> Please?

Sure, I am just not aware of your workshop scenario, could you merge both the flags together as my above suggestion. I will merge it right away.

henrygab:
> ... could you merge both the flags together as my above suggestion. I will merge it right away.

Hi @hathach,

If I were to merge both the flags, then it would not be possible to easily enable SysView without also enabling high debug output level.  In fact, it would be really difficult to enable a menu option that changes SysView without also changing debug level.  To enable the use of the custom menu options, it's necessary to have two distinct flags, else there would be a need to add exponentially larger menu options.

<details><summary>[Click to expand] An explanation why SysView only at debug level 3+ is problematic...</summary><P>

This issue is known in the software industry as a [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug), which is a combination of the term [Heisenburg](https://en.wikipedia.org/wiki/Uncertainty_principle) (the uncertainty principle) and the term bug.

Basically, a Heisenbug is a bug that, when you attempt to study it, disappears or changes its behavior.  Heisenbugs are often caused by subtle timing issues.  Here, when setting debug level to 3, there is a **_LOT_** more debug output that is being pushed through the serial connection.  This changes the execution timing of sketches relative to the real-time hardware peripherals.

Therefore, tracking a Heisenbug may require targeted changes to code, or removing the cause of as much timing variance as possible, while still logging the necessary information.  Being able to enable SysView while having zero serial port debug output is beneficial, because it means only the SysView logging overhead exists (which is relatively small).  This not only improves the ability to analyze timing-sensitive bugs such as Heisenbugs, but also allows for much more accurate analysis of where the board is spending its time.

</details>

In summary, having SysView tied to debug level is problematic.  There is a need to have two distinct variables for the two distinct options.

libraries/Adafruit_LittleFS/src/littlefs/lfs_util.h
```diff
@@ -56,21 +56,21 @@ extern "C"
 // Logging functions
 #ifndef LFS_NO_DEBUG
 #define LFS_DEBUG(fmt, ...) \
-    printf("lfs debug:%d: " fmt "\n", __LINE__, __VA_ARGS__)
+    PRINTF("lfs debug:%d: " fmt "\n", __LINE__, __VA_ARGS__)
```
hathach: @henrygab I got an local compilation issue with PRINTF is not defined when debug is enabled (which sneak through ci), it makes sense since lfs_util doens't include the `common_func.h` and it shouldn't though since it is an external library.

hathach: #467 revert this change until fixed probably later on.

Issue #462, opened 2020-03-19
Regex for arm-none-eabi-size output fails in arduino-cdt


toxynoid:
The regex for calculating the size with arm-none-eabi-size in platform.txt seem wrong to me
`recipe.size.regex=\.text\s+([0-9]+).*`
should be more
`recipe.size.regex=(?:\.text|\.data)\s+([0-9]+).*`
because as mentioned [here](https://mcuoneclipse.com/2013/04/14/text-data-and-bss-code-and-data-size-explained/) data is in the ROM as well. The non capturing regex group because the grouping match with index 1 gets evaluated in arduinocdt and there is only one capturing group in the regex for platform.txt in the arduino package as well.
This: 
`recipe.size.regex.data=(\.data|\.bss|\.stack)\s+([0-9]+).*`
should be:
`recipe.size.regex.data=(?:\.data|\.bss)\s+([0-9]+).*`
There is no line "stack" in the output of gcc size and again only second group should do capturing.

**Set up**
  - Adafruit nRF52 Boards Feather Express
  - 0.18.5
  - Ubuntu 19.10

hathach:
thanks for the issue, I am also long-time eclipse user, though I haven't used arduino cdt yet. Is it part of the cdt extension package. If you could give me brief on how to install it, I will try it out for confirmation

![image](https://user-images.githubusercontent.com/249515/77141609-1d80f380-6ab0-11ea-99a5-b6a52baf1d76.png)

toxynoid:
Am 20.03.20 um 07:38 schrieb Ha Thach:
> thanks for the issue, I am also long-time eclipse user, though I haven't
> used arduino cdt yet. Is it part of the cdt extension package. If you
> could give me brief on how to install it, I will try it out for
confirmation

Just open
"help -> eclipse market place"
and install
"Eclipse C++ IDE for Arduino 3.0"
After restart you have
"help -> Arduino Downloads Manager"
from there you add the platform see screenshot arduino-dl-manager.png.
Then after you created a new arduino project right click project and
activate the libraries. See libs.png

The recipe.objcopy.zip.pattern from platform.txt does not get executed
by cdt yet. I'm planning on implementing it.

Right now zip task is planned as Makefile target but it seems that the
zip task would be better suited for the "tools" section of platform.txt.
So I'm not sure which to use.

Do you have further specifications how platform.txt is to be
edited/interpreted?

Best

hathach:
[...]

PR #464, opened 2020-03-19
Update platform.txt

toxynoid:
See https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/462

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -112,8 +112,8 @@ recipe.output.save_file_hexu={build.project_name}.save.hex
 
 ## Compute size
 recipe.size.pattern="{compiler.path}{compiler.size.cmd}" -A "{build.path}/{build.project_name}.elf"
-recipe.size.regex=\.text\s+([0-9]+).*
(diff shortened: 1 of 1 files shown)


platform.txt
```diff
@@ -112,8 +112,8 @@ recipe.output.save_file_hexu={build.project_name}.save.hex
 
 ## Compute size
 recipe.size.pattern="{compiler.path}{compiler.size.cmd}" -A "{build.path}/{build.project_name}.elf"
-recipe.size.regex=\.text\s+([0-9]+).*
-recipe.size.regex.data=(\.data|\.bss|\.stack)\s+([0-9]+).*
+recipe.size.regex=(?:\.text|.\data)\s+([0-9]+).*
```
hathach: thanks for the PR, It is probably even better if you could add the `^` (size and datasize) so that we are exactly the same as Arduino samd core upstream here 
https://github.com/arduino/ArduinoCore-samd/blob/master/platform.txt#L120

hathach:
@toxynoid why closing the PR, you can keep modifying and push, the PR will get updated.

PR #467, opened 2020-03-20
partial revert PRITNF in lfs_util.h due to compilation error

hathach:
Follow up to #454 https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/454/files#r395434902
cause compilation error with debug level 1.


henrygab:
👀  I am sorry that slipped through.  Thanks!

PR #469, opened 2020-03-20
[WIP] Enable/Disable SysView via CFG_SYSVIEW

henrygab:
Of course, this commit ensures the following:

* Zero change to BSP defaults
* Zero change to compiled code
* Zero change to UI / menu options

### DIFF
--- a/cores/nRF5/freertos/config/FreeRTOSConfig.h
+++ b/cores/nRF5/freertos/config/FreeRTOSConfig.h
@@ -196,8 +196,8 @@ standard names - or at least those used in the unmodified vector table. */
      */
 #define configUSE_DISABLE_TICK_AUTO_CORRECTION_DEBUG     0
 
-// Sysview require at least debug level 3
-#if CFG_DEBUG >= 3
+// Sysview is enabled by default at debug level 3
+#if CFG_SYSVIEW
 #include "sysview/SEGGER_SYSVIEW_FreeRTOS.h"
 #endif
 
--- a/cores/nRF5/main.cpp
+++ b/cores/nRF5/main.cpp
@@ -24,8 +24,8 @@ void Bluefruit_printInfo() __attribute__((weak));
 void Bluefruit_printInfo() {}
 #endif
 
-// DEBUG Level 3
-#if CFG_DEBUG >= 3
+// From the UI, setting debug level to 3 will enable SysView
+#if CFG_SYSVIEW
 #include "SEGGER_SYSVIEW.h"
 #endif
 
@@ -80,7 +80,7 @@ int main( void )
   Adafruit_TinyUSB_Core_init();
 #endif
 
-#if CFG_DEBUG >= 3
+#if CFG_SYSVIEW
   SEGGER_SYSVIEW_Conf();
 #endif
 
--- a/boards.txt
+++ b/boards.txt
@@ -59,7 +59,8 @@ feather52832.menu.debug.l1.build.debug_flags=-DCFG_DEBUG=1
 feather52832.menu.debug.l2=Level 2 (Full Debug)
 feather52832.menu.debug.l2.build.debug_flags=-DCFG_DEBUG=2
 feather52832.menu.debug.l3=Level 3 (Segger SystemView)
-feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3 -DCFG_SYSVIEW=1
+feather52832.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3
+feather52832.menu.debug.l3.build.sysview_flags=-DCFG_SYSVIEW=1
 
 # ----------------------------------
 # Bluefruit Feather nRF52840 Express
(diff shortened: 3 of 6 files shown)


cores/nRF5/main.cpp
```diff
@@ -24,8 +24,8 @@ void Bluefruit_printInfo() __attribute__((weak));
 void Bluefruit_printInfo() {}
 #endif
 
-// DEBUG Level 3
-#if CFG_DEBUG >= 3
+// From the UI, setting debug level to 3 will enable SysView
```
hathach: the comment text is not relevant anymore, you can safely remove it

cores/nRF5/freertos/config/FreeRTOSConfig.h
```diff
@@ -196,8 +196,8 @@ standard names - or at least those used in the unmodified vector table. */
      */
 #define configUSE_DISABLE_TICK_AUTO_CORRECTION_DEBUG     0
 
-// Sysview require at least debug level 3
-#if CFG_DEBUG >= 3
+// Sysview is enabled by default at debug level 3
```
hathach: the comment text is not relevant anymore, you can safely remove it

cores/nRF5/main.cpp
henrygab: I think this comment is very accurate.  When the debug level is set to 3, SysView is enabled.  This is the old behavior, and continues to be the behavior with this PR.  So, I am not sure I understand why this would not be a relevant comment?

cores/nRF5/freertos/config/FreeRTOSConfig.h
henrygab: <details><summary>Same as prior response...</summary><P>

[Prior response](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/469#discussion_r397044991) said the following:

> I think this comment is very accurate. When the debug level is set to 3, SysView is enabled. This is the old behavior, and continues to be the behavior with this PR. So, I am not sure I understand why this would not be a relevant comment?

</details>

hathach:
Thanks @henrygab for detail explanation as usual, I just only want to make sure the CFLAGS passed to compiler doesn't define `-DCFG_SYSVIEW` **twice**
- once with 0 by `build.sysview_flags` and 
- one with the debug level 3 menu

henrygab:
looking into this.  I had not seen a problem, but now I get warnings about exactly this, and yes, it shows up twice on the build line.

I will update title and tag you again when I know what difference (vs the old branch) is causing this, and the minimal change.

cores/nRF5/freertos/config/FreeRTOSConfig.h
hathach: @henrygab yeah, previously the text is meant to explain why we enable sysview when DEBUG >= 3, now sysvew has its own macro which is self described already. So it is safe to remove, though it is not really an important at all.

henrygab:
The problem is now clear.
* I defined the two distinct variables in platforms.txt.
* I originally had separate menu items to set debug vs. set sysview
* I even provided a correct example in boards.local.txt snippet, above
* But, when I combined the menu entries, I squished both `#defines` into `build.debug_flags`, rather than splitting the sysview define into `build.sysview_flags`

In other words, I was seeing what I intended, not what I wrote.

The good news is that the fix was straightforward.

@hathach  -- **_Thank you_** for catching this, and for your patience with me as I failed to see what you were pointing me at, and which  was right there in front of me the whole time.

henrygab:
@hathach -- I had prefixed the title with "[WIP]" tag (work in progress).

**_No problem in this case_**, because the merged commit was the correct fix.
<details><summary>[expand] tested matrix</summary><P>

I had manually validated build logs from compilation of the little_fs stress test sketch, for each of the following configurations:
* Feather nRF52832
  * Debug 0, 1, 2, 3
* Feather Express nRF52840
  * Debug 0, 1, 2, 3
  * Platform.local.txt forcing sysview on, debug 0
  * Boards.local.txt forcing sysview on, debug 1
  * Boards.local.txt forcing sysview off, debug 3

As expected, it's not possible to force-disable sysview from platform.local.txt, when set to debug level 3.  This is because platform.local.txt is applied **_before_** the boards.txt settings, and thus the boards.txt settings for debug level 3 will enable sysview.  Of course, if that's the desire, it can be overridden in boards.local.txt.

</details><HR/>

Did you know the fix I needed to make, and thus felt confident with that commit?

If not, is there a different method preferred to mark a PR as "not ready / work in progress"?

_I ask only to help things continue to work smoothly for the future._

hathach:
@henrygab when submit an PR, you can choose to submit a draft PR
![image[https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

for case with debug level 3 + the use of platform.local.txt. That is OK, people normally won't use the local file, those used it will find a way to figure it out.

henrygab:
Hmmm... there doesn't appear to be a way to convert a PR to a draft PR:

According to the folks on [stackoverflow](https://stackoverflow.com/questions/55070723/how-do-i-change-from-a-pr-to-a-draft-pr-at-github), the most common workaround is to have the title contain `WIP`.  In fact, gitlab uses this to swap a PR from draft/ready and back.

So, I guess I will just close PRs if they ever reach non-draft state and then require changes.  Too bad, as it loses context from prior conversations.

PR #472, opened 2020-03-23
Update platform.txt

toxynoid:
As recommended

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -115,8 +115,8 @@ recipe.output.save_file_hexu={build.project_name}.save.hex
 
 ## Compute size
 recipe.size.pattern="{compiler.path}{compiler.size.cmd}" -A "{build.path}/{build.project_name}.elf"
-recipe.size.regex=^(\.text|.\data)\s+([0-9]+).*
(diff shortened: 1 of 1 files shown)


platform.txt
```diff
@@ -115,8 +115,8 @@ recipe.output.save_file_hexu={build.project_name}.save.hex
 
 ## Compute size
 recipe.size.pattern="{compiler.path}{compiler.size.cmd}" -A "{build.path}/{build.project_name}.elf"
-recipe.size.regex=\.text\s+([0-9]+).*
-recipe.size.regex.data=(\.data|\.bss|\.stack)\s+([0-9]+).*
+recipe.size.regex=^(\.text|.\data)\s+([0-9]+).*
```
hathach: sorry I would have said clearly that the regex should be like the https://github.com/arduino/ArduinoCore-samd/blob/master/platform.txt#L120 which is 

```
recipe.size.regex=^(?:\.text|\.data|)\s+([0-9]+).*
recipe.size.regex.data=^(?:\.data|\.bss)\s+([0-9]+).*
```

You don't have to close the PR, just keep commit and pushing to your patch-5 branch, it will update the PR as well. Sorry for any inconveniences.

PR #482, opened 2020-04-13
Added Missing Pin Assignments for Xenon Board

Montvydas:
Xenon board can tell when it is getting powered through the USB rather than a battery and when a battery is being charged. Added required pins inside variant files and the pins work as specified:

- When PIN_PWD is HIGH, Xenon is being powered through the USB
- When PIN_CHG is LOW, a battery attached to Xenon is being charged

Note: Prior to using these pins it is neccessary to specify pinMode(PIN_CHG, INPUT) and pinMode(PIN_PWR, INPUT) due to how they're being implemented in hardware. Using these two pins 3 states of power can be deduced:
- hasUsbPower = digitalRead(PIN_PWR);
- isCharging = (hasUsbPower && !digitalRead(PIN_CHG));
- onBatteryPower = !hasUsbPower;

BUTTON_FRESET pin number assignment was fixed as previously it was pointing to a wrong pin.

Added the missing LED1 initialisation, which now mimics the behaviour of the other nRF52 boards.

### DIFF
--- a/variants/particle_xenon/variant.cpp
+++ b/variants/particle_xenon/variant.cpp
@@ -52,30 +52,42 @@ const uint32_t g_ADigitalPinMap[] =
   _PINNUM(0, 11),   // P0.11 (MODE)
   _PINNUM(0, 18),   // P0.18 (RESET)
 
+  // D22 .. D24
   // LEDS
   _PINNUM(0, 13),   // P0.13 (RGB_RED)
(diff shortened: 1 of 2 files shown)


Montvydas:
Thanks for taking such a quick look!

hathach:
I am not normally that quick (normally scratching head debuggin) :D

PR #484, opened 2020-04-15
Flipped SDA and SCL

kattemjau:
Firmware is now according to PCB layout and marking on Adafruit NRF52840 Express. This easy fix removed the need for a redesign!


dhalbert:
What bug is this fixing? `SCL` is `P0.11` and `SDA` is `P0.12`, by the [schematic](https://learn.adafruit.com/assets/68545), and that mapping works. Are you expecting D22 an D23 to be in a certain order?

kattemjau:
Hi, you are right. It was a long day yesterday and the issue was somewhere in my schematic rather then the reference design. Sorry for wasting your time. Please reject this commit!

PR #485, opened 2020-04-18
GPIOTE fixes for attachInterrupt() / detachInterrupt()

### LINKED ISSUE #477: GPIOTE channels are overwritten if in use by library
**Describe the bug**
If a library uses a GPIOTE channel 0, and then the sketch calls attachInterrupt(), the library's configuration is corrupted.

  - nRF52840 Feather Express
  - BSP from commit 0019306c (0.19.0)
  - Windows 10
  - Log: N/A

**To Reproduce**
1. Configure `GPIOTE` channel 0 and `PPI` channels 0-4 (e.g., as a user, using an opaque library) to configure GPIO output state changes without CPU involvement
2. Once that sketch is working, add button on another GPIO, using `attachInterrupt()` to register an interrupt handler for that button
3. Sadness, as things stop working
4. Comment out call to `attachInterrupt()`
5. Happiness, as sketch starts working again!
6. Track down underlying issue, and realize the cause is similar to #459 and #447

**Expected behavior**
BSP doesn't use peripherals that are clearly already in use.

henrygab:
# WORK IN PROGRESS [WIP]
Fix  #477

Although the bug is confirmed (see [gist](https://gist.github.com/henrygab/d02e6b03be5636d51fc0fe8748cb3b21)), the fix is not fully ready.

Although the concept is clear from the first two commits in this branch, there is still at least one bug that prevents integration at the moment.

tagging myself @henrygab

### DIFF
--- a/cores/nRF5/WInterrupts.c
+++ b/cores/nRF5/WInterrupts.c
@@ -132,7 +132,7 @@ int attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
     // skip if already used in AttachInterrupt for another pin
     if (channelMap[ch] != -1) continue;
     // skip if channel is not disabled (e.g., in use by some other component or library)
-    if (nrf_gpiote_te_is_enabled(ch)) continue;
+    if (nrf_gpiote_te_is_enabled(NRF_GPIOTE, ch)) continue;
     uint32_t tmp = NRF_GPIOTE->CONFIG[ch];
     tmp &= oldRegMask;
     tmp |= newRegBits;
--- a/cores/nRF5/WInterrupts.c
+++ b/cores/nRF5/WInterrupts.c
@@ -107,57 +107,52 @@ int attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
     ((polarity                 << GPIOTE_CONFIG_POLARITY_Pos) & GPIOTE_CONFIG_POLARITY_Msk) |
     ((GPIOTE_CONFIG_MODE_Event << GPIOTE_CONFIG_MODE_Pos    ) & GPIOTE_CONFIG_MODE_Msk    ) ;
 
-  // To avoid unexpected consequences, first loop through the channelMap
-  // to preferably re-use any channel that was already in use (even if
-  // earlier channel is no longer in use).
-  for (int ch = 0; ch < NUMBER_OF_GPIO_TE; ch++) {
-    // skip if the channel was not already assigned to this pin
-    if ((uint32_t)channelMap[ch] != pin) continue;
+  int ch = -1;
+  int newChannel = 0;
 
-    // The pin is already allocated in the channelMap
-    // update the polarity (when events fire) and callbacks
-    // However, do NOT clear any GPIOTE events
-    uint32_t tmp = NRF_GPIOTE->CONFIG[ch];
-    channelMap[ch] = pin;
-    callbacksInt[ch] = callback;
-    callbackDeferred[ch] = deferred;
-    tmp &= oldRegMask;
-    tmp |= newRegBits;
-    NRF_GPIOTE->CONFIG[ch] = tmp;
-    asm volatile ("" : : : "memory");
-    __asm__ __volatile__ ("nop\n\tnop\n\tnop\n\tnop\n");
-    NRF_GPIOTE->INTENSET = (1 << ch); // old code did this ... no harm in ensuring this is set
-    return (1 << ch);
+  // Find channel where pin is already assigned, if any
+  for (int i = 0; i < NUMBER_OF_GPIO_TE; i++) {
+    if ((uint32_t)channelMap[i] != pin) continue;
+    ch = i;
+    break;
+  }
+  // else, find one not already mapped and also not in use by others
+  if (ch == -1) {
+    for (int i = 0; i < NUMBER_OF_GPIO_TE; i++) {
+      if (channelMap[i] != -1) continue;
+      if (nrf_gpiote_te_is_enabled(NRF_GPIOTE, i)) continue;
+      
+      ch = i;
+      newChannel = 1;
+      break;
+    }
+  }
+  // if no channel found, exit
+  if (ch == -1) {
+    return 0; // no channel available
   }
 
-  // When the pin isn't already configured for interrupts, then attempt to
-  // find an unused GPIOTE channel.  This code is identical to the above,
-  // except for also validating that the MODE bits show the channel is disabled.
-  for (int ch=0; ch < NUMBER_OF_GPIO_TE; ch++) {
-    // skip if already used in AttachInterrupt for another pin
-    if (channelMap[ch] != -1) continue;
-    // skip if channel is not disabled (e.g., in use by some other component or library)
-    if (nrf_gpiote_te_is_enabled(NRF_GPIOTE, ch)) continue;
+  uint32_t tmp = NRF_GPIOTE->CONFIG[ch];
+  tmp &= oldRegMask;
+  tmp |= newRegBits;                 // for existing channel, effectively updates only the polarity
+  channelMap[ch]         = pin;      // harmless for existing channel
+  callbacksInt[ch]       = callback; // caller might be updating this for existing channel
+  callbackDeferred[ch]   = deferred; // caller might be updating this for existing channel
+  NRF_GPIOTE->CONFIG[ch] = tmp;
 
-    // clear any old events on this GPIOTE channel
-    uint32_t tmp = NRF_GPIOTE->CONFIG[ch];
-    channelMap[ch] = pin;
-    callbacksInt[ch] = callback;
-    callbackDeferred[ch] = deferred;
-    tmp &= oldRegMask;
-    tmp |= newRegBits;
-    // TODO: make check/set for new channel an atomic operation
-    NRF_GPIOTE->CONFIG[ch] = tmp;
-    asm volatile ("" : : : "memory");
-    __asm__ __volatile__ ("nop\n\tnop\n\tnop\n\tnop\n");
+  // For a new channel, additionally ensure no old events existed, and enable the interrupt
+  if (newChannel) {
+    // Why the memory barrier and NOPs? 
+    // See page 4 at https://web.archive.org/web/20190823222657/http://infocenter.arm.com/help/topic/com.arm.doc.dai0321a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf
+    // and recall the AHB runs at 16MHz, so may take 4 cycles before peripheral register is actually updated.
+    __DSB(); __NOP();__NOP();__NOP();__NOP();
     NRF_GPIOTE->EVENTS_IN[ch] = 0;
-    asm volatile ("" : : : "memory");
-    __asm__ __volatile__ ("nop\n\tnop\n\tnop\n\tnop\n");
+    __DSB(); __NOP();__NOP();__NOP();__NOP();
     NRF_GPIOTE->INTENSET = (1 << ch);
-    return (1 << ch);
   }
-  // Else, pin was neither already setup, nor could a GPIOTE be allocated
-  return 0;
+
+  // Finally, indicate to caller the allocated / updated channel
+  return (1 << ch);
 }
 
 /*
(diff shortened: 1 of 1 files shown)


henrygab:
This fixes the following issues:

1. Prevents `attachInterrupt()` from grabbing a GPIOTE channel if it's already in use by anything else (checks if the channel is enabled using [`nrf_gpiote_te_is_enabled()`](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R140)

2. Prevents `attachInterrupt()` from getting confused when an unused entry exists in [`channelMap`](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R115) by splitting loop into two logical searches:  first search for pin already in map, and second only search for an empty channel if none of the entries match the pin.

3. Reduces slow (across 16MHz bus) [reads](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R207-R209) for minor ISR perf improvement.

4. Add a few explicit "sequence points", to ensure GCC optimizations do not re-order memory accesses in violation of the code's intentions.

Tested and verified, stepping through the [example gist](https://gist.github.com/henrygab/d02e6b03be5636d51fc0fe8748cb3b21).

henrygab:
@hathach -- Ready for final review, this fixes the following issues:

1. Prevents `attachInterrupt()` from grabbing a GPIOTE channel if it's already in use by anything else (checks if the channel is enabled using [`nrf_gpiote_te_is_enabled()`](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R140)

2. Prevents `attachInterrupt()` from getting confused when an unused entry exists in [`channelMap`](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R115) by splitting loop into two logical searches:  first search for pin already in map, and second only search for an empty channel if none of the entries match the pin.

3. Reduces slow (across 16MHz bus) [reads](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485/files#diff-1e6265b634019f1a57124be4c2feeab3R207-R209) for minor ISR perf improvement.

4. Add a few explicit "sequence points", to ensure GCC optimizations do not re-order memory accesses in violation of the code's intentions.

Tested and verified, stepping through the [example gist](https://gist.github.com/henrygab/d02e6b03be5636d51fc0fe8748cb3b21).

cores/nRF5/WInterrupts.c
```diff
@@ -95,24 +96,67 @@ int attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
...
+  const uint32_t oldRegMask = ~(GPIOTE_CONFIG_PORT_PIN_Msk | GPIOTE_CONFIG_POLARITY_Msk | GPIOTE_CONFIG_MODE_Msk);
+  const uint32_t newRegBits =
+    ((pin                      << GPIOTE_CONFIG_PSEL_Pos    ) & GPIOTE_CONFIG_PORT_PIN_Msk) |
+    ((polarity                 << GPIOTE_CONFIG_POLARITY_Pos) & GPIOTE_CONFIG_POLARITY_Msk) |
+    ((GPIOTE_CONFIG_MODE_Event << GPIOTE_CONFIG_MODE_Pos    ) & GPIOTE_CONFIG_MODE_Msk    ) ;
+
+  // To avoid unexpected consequences, first loop through the channelMap
+  // to preferably re-use any channel that was already in use (even if
+  // earlier channel is no longer in use).
   for (int ch = 0; ch < NUMBER_OF_GPIO_TE; ch++) {
```
hathach, line 113: two loops approach is good, but It duplicates the actual handling code. I would be better to move the 2 loop() into a seperate find_available_channel() function, that
- return currently used channel if any else
- return first empty
- return (-1) if non

Then we will only need to write handling code once.

cores/nRF5/WInterrupts.c
```diff
@@ -148,30 +206,58 @@ void GPIOTE_IRQHandler()
...
-    volatile uint32_t dummy = *((volatile uint32_t *)((uint32_t)NRF_GPIOTE + event));
-    (void)dummy;
-#endif
     }
 
-    event = (uint32_t)((uint32_t)event + 4);
+    // clear the event
+    NRF_GPIOTE->EVENTS_IN[ch] = 0;
   }
+#if __CORTEX_M == 0x04
```
hathach, line 234: for this, we don't need to have paragraph length of comment, we can just put a comment line with a section quote from specs. Please remove the asm code and use either one of  __ISB __DSB __DMB as you see approriate.
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/cmsis/include/cmsis_gcc.h#L438

henrygab: OK, reducing comments and changing to instrinsics.  Will push soon and tag you then.

cores/nRF5/WInterrupts.c
henrygab: Well, the handling code is **_not_** common (and should not be common) between the two cases.  Still, I am updating per your request.

henrygab:
@hathach -- As requested, and ready for review.

* all assembly code is removed and replaced with intrinsics.
* comments are greatly reduced.
* configuration of channels in `attachInterrupt()` is now using common configuration code block

cores/nRF5/WInterrupts.c
```diff
@@ -95,25 +96,63 @@ int attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
...
+  callbacksInt[ch]       = callback; // caller might be updating this for existing channel
+  callbackDeferred[ch]   = deferred; // caller might be updating this for existing channel
+  NRF_GPIOTE->CONFIG[ch] = tmp;
+
+  // For a new channel, additionally ensure no old events existed, and enable the interrupt
+  if (newChannel) {
+    // Why the memory barrier and NOPs? 
+    // See page 4 at https://web.archive.org/web/20190823222657/http://infocenter.arm.com/help/topic/com.arm.doc.dai0321a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf
+    // and recall the AHB runs at 16MHz, so may take 4 cycles before peripheral register is actually updated.
+    __DSB(); __NOP();__NOP();__NOP();__NOP();
```
hathach, line 148: please remove these _DSB(), the order of these 2 are not important. Plus I don't see the reason why gcc would run it differently.

cores/nRF5/WInterrupts.c
```diff
@@ -129,14 +168,19 @@ void detachInterrupt(uint32_t pin)
 
   for (int ch = 0; ch < NUMBER_OF_GPIO_TE; ch++) {
     if ((uint32_t)channelMap[ch] == pin) {
+      // Why the memory barrier and NOPs? 
+      // See page 4 at https://web.archive.org/web/20190823222657/http://infocenter.arm.com/help/topic/com.arm.doc.dai0321a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf
+      // and recall the AHB runs at 16MHz, so may take 4 cycles before peripheral register is actually updated.
+      NRF_GPIOTE->INTENCLR = (1 << ch);
+      __DSB(); __NOP();__NOP();__NOP();__NOP();
```
hathach, line 175: please remove these _DSB(), the order of these 2 are not important. Plus I don't see the reason why gcc would run it differently.

cores/nRF5/WInterrupts.c
```diff
@@ -95,25 +96,63 @@ int attachInterrupt(uint32_t pin, voidFuncPtr callback, uint32_t mode)
...
-
-      NRF_GPIOTE->INTENSET = (1 << ch);
-
-      return (1 << ch);
+  // All information for the configuration is known, except the prior values
+  // of the config register.  Pre-compute the mask and new bits for later use.
+  //     CONFIG[n] = (CONFIG[n] & oldRegMask) | newRegBits;
+  //
+  // Three fields are configured here: PORT/PIN, POLARITY, MODE
+  const uint32_t oldRegMask = ~(GPIOTE_CONFIG_PORT_PIN_Msk | GPIOTE_CONFIG_POLARITY_Msk | GPIOTE_CONFIG_MODE_Msk);
```
hathach, line 104: maybe move it near its usage. feel free to leave it as it is. I could do it later on.

cores/nRF5/WInterrupts.c
```diff
@@ -148,30 +192,39 @@ void GPIOTE_IRQHandler()
...
 
-    event = (uint32_t)((uint32_t)event + 4);
+    // clear the event
+    NRF_GPIOTE->EVENTS_IN[ch] = 0;
   }
+#if __CORTEX_M == 0x04
+  // See note at nRF52840_PS_v1.1.pdf section 6.1.8 ("interrupt clearing")
+  // See also https://gcc.gnu.org/onlinedocs/gcc/Volatiles.html for why
+  // using memory barrier instead of read of an unrelated volatile
+  __DSB(); __NOP();__NOP();__NOP();__NOP();
```
hathach, line 224: thanks, this is much better than the original read then discard register. I think this is the only place need adding the _DSB()

cores/nRF5/WInterrupts.c
henrygab: sure.

henrygab:
> Thank you very much @henrygab for putting lots of effort into the PR. Though I think we only need to do _DSB() at the isr only. Other places are not really needed. I am open to any discussions.

First, thanks for the detailed reviews.  As for the above, I am not sure that it is safe to exclude these _DSB().  Perhaps a simplified example:

```C++
typedef struct { // simplified version
  volatile uint32_t  EVENTS_IN;
  volatile uint32_t  INTENSET;
  volatile uint32_t  INTENCLR;
  volatile uint32_t  CONFIG;
} NRF_GPIOTE_Type;

void setup(void) {
    gpiote->CONFIG = 1; // if left at old value, result is interrupt storm
    // _DMB() omitted ???
    gpiote->INTENSET = 1;  
}
```

It appears legal for either the compiler, or the CPU internally, to re-order these two writes, as they are not marked as dependent upon each other.  Thus, could this have the actual order?
```
    gpiote->INTENSET = 1;
    gpiote->CONFIG = 1;
```

_(caused by them being different and unrelated addresses)_

Is there any guarantee of either or  both the following?

1. That the compiler would never re-ordering volatile memory accesses to distinct, non-dependent locations

2. That the CPU would never re-order volatile memory accesses to distinct, non-dependent locations

hathach:
This is not an example, it is a theory, a example should make gcc generate code that execute in reverse order. If gcc can randomly re-order every register write, we will have to just add _DSB() on every register access function. But we don't see that even within Nordic nrfx or cmsis driver. It is not necessary, furthermore in this case CONFIG and INTSET/INTCLR can just be executed in any order and sill wouldn't affect the behavior.

henrygab:
@hathach -- Thank you for your patience, and kind words.

<details><summary>We agree that this particular CPU does not require a memory barrier.</summary><P/>

The nRF52840 is a Cortex-M processor.  ARM explicitly guarantees that **_the CPU_** will not reorder the operations in this case.  Page 20 of the [ARM document](https://web.archive.org/web/20190823222657/http://infocenter.arm.com/help/topic/com.arm.doc.dai0321a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf) says:

> ### 4.2 Device (Peripheral) access
> There is no need to use a memory barrier instruction between each step during peripheral
programming or peripheral access:
> * Architecturally: Access to the same device must be in the programmed order.
> * Implementation: Cortex-M processors do not reorder data transfers

Thus, this particular CPU does not require them.  The only remaining thing to ensure is that the compiler does not require at least a compiler barrier (see [https://web.archive.org/web/20190823124518/https://blog.regehr.org/archives/28], item 5, for explanation).

</details>

Regardless, I will remove the explicit sequence points / memory barriers in the programming.  Be careful, however, if moving this code to any other compiler version or CPU architecture, as the code might not be safely portable.

cores/nRF5/WInterrupts.c
henrygab: The order is definitely important!  If interrupts are enabled for a device that is not using the correct configuration, it can range from not working, to spurious interrupts, to hard-lock (interrupt storm).

That said, I've removed them ... but advise caution if moving to a different compiler or chip architecture (including multi-core)

cores/nRF5/WInterrupts.c
henrygab: see response to [next comment](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/485#discussion_r416054396)

PR #493, opened 2020-04-26
Fix off-by-one in shiftIn function

dzervas:
This fixes the common `++i` vs `i++` bug in the `shiftIn` function of the Arduino core.

Took me a whole day to find... :cry:


dzervas:
this is invalid, please ignore

hathach:
no problems, keep PRs coming :)

Issue #124, opened 2018-02-26
support pairing using LESC procedure 


hathach:
LESC use Elliptic Curve Diffie Hellman (ECDH) for key generation, there it requires crypto library with ARM Cryptocell CC310 on nrf52840

https://infocenter.nordicsemi.com/index.jsp?topic=%2Fsdk_nrf5_v16.0.0%2Flib_cryptocell.html&resultof=%22nrf_cc310%22

PR #494, opened 2020-04-30
Making the nRF52840 Express BSP multiprotocol friendly

orrmany:
Hi, I would like to ask for some modifications to the nRF52 Arduino core for making it easier to use nRF52840 Express with BLE & ANT+ protocol stacks together. That necessitates 
- modificiations to the bootloader (which will come in a separate PR, since that is a different repo)
- modifications to board menu (I added a new board menu for the board with the S340 SoftDevice (SD) calling it "Bluefruit Feather nRF52840 Express s340")
- some modification to the class AdafruitBluefruit , since the ANT+ protocol stack needs its own FreeRTOS task and the S340 soft-device-enable() method has longer method-signature. 

For multiprotocol usage I had to make the class AdafruitBluefruit a little better behaving FreeRTOS citizen. The changes are as follows: 
- If the ANT_LICENSE_KEY symbol is defined then the code assumes S340 softdevice and it calls sd_softdevice_enable(...) with 3 args, otherwise it falls back to the S140 variant of 2 args.
- I added a private pointer to an optional ANT+ task handler semaphore, which is initialized to NULL and a corresponding setter method to supply an external semaphore reference
-  SD_EVT_IRQHandler() now checks that semaphore pointer and if it is set then it throws a "give" to the new semaphore, as well, beyond the legacy BLE and the SoC semaphores.

With these changes I was able to run the BLE & ANT stacks concurrently on my nRF52840 Express updated with S340 SD. I am currently using some ANT+ sensor reading together with Adafruit's BLEUART example, which latter needed no changes at all. 

As such, I believe all these changes are backward compatible for boards with S140. The changes are transparent for legacy Bluefruit52-dependent code, i.e., these should be harmless for existing users' sketches. The only visible difference for the users would be the appearance of a new board-type in the Arduino/PlatformIo board-menu. 

You can read further details at my Wiki below (Note, those articles were written a few months ago, so they reflect an earlier SW version of your code): 
- [S340 SoftDevice & Adafruit nRF52840 Feather Express](http://orrmany.hu/wiki/doku.php?id=arduino:s340_softdevice_adafruit_nrf52840_feather_express)
- [Arduino IDE integration for the nRF52840 Feather Express with S340](http://orrmany.hu/wiki/doku.php?id=s340:arduino_ide_integration_for_the_nrf52840_feather_express_with_s340)
- [PlatformIO IDE integration for the nRF52840 Feather Express with S340](http://orrmany.hu/wiki/doku.php?id=s340:platformio_ide_integration_for_the_nrf52840_feather_express_with_s340)

Credit shall also go to @rtgree01 ([Ryan Green](https://github.com/rtgree01/Adafruit_nRF52_Arduino)) who made most of the early hacks of these and he kindly shared his results with me and @cujomalainey [Curtis Malainey](https://github.com/cujomalainey).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,7 @@
 # Ignore local overrides of platform.txt and boards.txt,
 /boards.local.txt
 /platform.local.txt
+
+# Ignore S340 SoftDevice due to license restrictions
 bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.*
 cores/nRF5/nordic/softdevice/s340_nrf52_6.1.1_API/*
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,5 @@
 # Ignore local overrides of platform.txt and boards.txt,
 /boards.local.txt
 /platform.local.txt
-bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.hex
-bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.zip
+bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.*
 cores/nRF5/nordic/softdevice/s340_nrf52_6.1.1_API/*
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@
 # Ignore local overrides of platform.txt and boards.txt,
 /boards.local.txt
 /platform.local.txt
+bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.hex
+bootloader/feather_nrf52840_express/feather_nrf52840_express_bootloader-0.3.0_s340_6.1.1.zip
+cores/nRF5/nordic/softdevice/s340_nrf52_6.1.1_API/*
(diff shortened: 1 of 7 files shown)


orrmany:
The corresponding [PR #124 for the Bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/124) has been submitted.

orrmany:
Please note, that altough this PR is related to [PR #124 for the Bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader/pull/124), but they are independent. This PR can be accepted, IMHO, independently from the other PR. 
Of course one would need to add the S340 files here, as well, for the new board type to work, but all the _other_ board types are supposed to work just fine with, or without the S340 files.

hathach:
Thank you for your PR, unfortunately we currently have no interest on ANT protocol due to its licencing restriction. The PR does not have much impact to the code base, however, the present of Feather nRF52840 with S340 can cause people to think that we support ANT protocol and can lead to series of unnecessary support questions/tickets. And I personally don't like to have code that we couldn't run or tested ourself since Softdevice and its API will surely continue to change. 
 
Since it is such an easy modification, I would suggest you to keep a folk or provide an guide for people to do so. There a lots of other steps such as upgrading bootloader, especially it needs its own ANT library to make use of the protocol anyway.

orrmany:
@hathach: this request is not just  ANT specific. 
Would another PR just about the changes to bluefruit.cpp and bluefruit.h more acceptable?

hathach:
we may merge minimal changes to bluefruit files provided there is absolutely no added code when ANT is not defined.

orrmany:
@hathach:  so as a second attempt another pull request with reduced (Bluefruit52 only) scope is created instead: "Making the nRF52840 Express BSP multiprotocol friendly #494"
Please kindly consider that

orrmany:
@hathach: so as a second attempt another pull request with reduced (Bluefruit52 only) scope is created instead: "Develop multiprotocol ble #501"
Please kindly consider that

PR #499, opened 2020-05-07
type-safe array macros

### LINKED ISSUE #498: Headers define non-type-safe array macros
**Describe the bug**
The headers defines macros that require array parameters, but which are not type-safe.
This is a [well-known problem](https://stackoverflow.com/questions/19452971/array-size-macro-that-rejects-pointers), especially difficult to do correctly for C99.

  - BSP 0.20.1

**To Reproduce**
1. This shows how the `arrcount()` and `arrclr` macro can be used **_unsafely_**.  This diff inserts into a `.C` file, while the same code can go into a sketch (which shows the problem in C++ source files):

```diff
diff --git a/cores/nRF5/pulse.c b/cores/nRF5/pulse.c
index f930a980..c438bf50 100644
--- a/cores/nRF5/pulse.c
+++ b/cores/nRF5/pulse.c
@@ -24,12 +24,22 @@
 extern unsigned long countPulseASM(const volatile uint32_t *port, uint32_t bit, uint32_t stateMask, unsigned long maxloops);

 unsigned long countPulseASM(const volatile uint32_t *port, uint32_t bit, uint32_t stateMask, unsigned long maxloops)
 {
   unsigned long width = 0;

+  uint16_t array[40];
+  size_t i;
+  uint16_t * pointer = array;
+  for (i = 0; i < arrcount(pointer); i++ );
+  arrclr(pointer);
+  for (i = 0; i < arrcount(array); i++ );
+  arrclr(array);
+
+
```

2. Compile the sketch ... notice that no warnings or failures are emitted for the improper use of a pointer in those macros ... which makes them **_dangerous_**.

**Expected behavior**
Use of macros that ensure parameters are arrays, when that is required for proper behavior of the macro.

henrygab:
Fix #498 ... add type-safety for **_both_** C and C++

### DIFF
--- a/cores/nRF5/common_func.h
+++ b/cores/nRF5/common_func.h
@@ -37,17 +37,47 @@
 #ifndef _COMMON_FUNC_H_
 #define _COMMON_FUNC_H_
 
+
 #ifdef __cplusplus
- extern "C" {
+  namespace ADAFRUIT_DETAIL
+  {
+      template <typename T, size_t N>
+      constexpr size_t arrcount_fails_if_not_array(T const (&)[N]) noexcept
+      {
+          return N;
+      }
+  }
+  extern "C" {
+  #define arrcount(arr) ADAFRUIT_DETAIL::arrcount_fails_if_not_array(arr)
+  #define ADA_STATIC_ASSERT(const_expr, message)   static_assert(const_expr, message)
+#elif __STDC_VERSION__ >= 201112L
+  // Even C can have type-safety for equivalent of ARRAY_SIZE() macro, when using GCC (which this BSP does)
+  #define __ADA_COMPATIBLE_TYPES(a,b)   __builtin_types_compatible_p(typeof(a), typeof(b)) // GCC extensions
+  #define __ADA_BUILD_ERROR_ON_ZERO(x)  (sizeof(struct { int:-!!(x)*0x1ee7;})) // if x is zero, reports "error: negative width in bit-field '<anonymous>'"
+  #define __ADA_MUST_BE_ARRAY(x)        __ADA_BUILD_ERROR_ON_ZERO(__ADA_COMPATIBLE_TYPES((x), &(*x)))
+  #define ADA_STATIC_ASSERT(const_expr, message)   _Static_assert(const_expr, message)
+  #define arrcount(_arr)       ( (sizeof(_arr) / sizeof((_arr)[0])) + __ADA_MUST_BE_ARRAY(_arr) ) // compile-time error if not an array
+#else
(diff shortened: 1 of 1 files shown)


henrygab:
@hathach -- Simple concept here.  Basically, this causes improper use of the `ARRAY_SIZE()` / `NELEMENTS()` / `arrcount()` style macros to be a compilation-time bug, rather than silently doing the wrong thing.

FYI, I tested against both C and C++ source files, both with proper and improper usage of the `arrcount()` and `arrclr()` macros.  (i.e., used with local array == success, used with a pointer == failure).

Let me know if any questions.

hathach:
@henrygab hihi, sorry,  am still busy with bootloader a bit, will review asap. Thanks :)

henrygab:
Keeping this thread current also.

henrygab:
Keeping this thread current.

cores/nRF5/common_func.h
```diff
@@ -37,17 +37,48 @@
 #ifndef _COMMON_FUNC_H_
 #define _COMMON_FUNC_H_
 
+
 #ifdef __cplusplus
- extern "C" {
+  // namespace and templates must be outside the `extern "C"` declaration...
+  namespace ADAFRUIT_DETAIL
```
hathach, line 43: Thank you for you patient, I should have done this long ago. Since I find this very interesting and like to do a bit of research myself, but couldn't find time. So yeah, it is not really the reason of delay but worth mentionning :sweat_smile: 

I like the C-only version more than the C++ version, can be just use the C-only version for this. If not then I am happy to merge it.

henrygab: Hi @hathach,

Thanks for the review.  

It has been some time since I wrote this.  I have tested only the C++ version when compiling C++ source files, and have tested only the C version when compiling C source files.  Also, the C version requires GCC extensions, while the C++ version is standards-compliant.

One large benefit of separate C vs C++ versions is that the C++ version results in more clear error messages.  After all, an error message of `"error: negative width in bit-field '<anonymous>'"` is not really clear as to the cause.  In contrast, an error message that includes the function name `arrcount_fails_if_not_array`, is substantially more clear as to the problem.

Of course, if you are happy with the C version error message, then please just confirm this is your preference, and I will adjust lines 41 and 54 accordingly.

then simply exclude the same as I think the C version is less capable than the C++ version ... for example, it requires GCC extensions, which perhaps not all compilers support.

However, please feel free to modify this branch in any way you see fit.  If the C version works in all cases, and you are happy with the

hathach: @henrygab thanks for clarification, I merge it now.

PR #501, opened 2020-05-07
Make Bluferuit52 multiprotocol friendly

orrmany:
This pull request is to make Bluefruit52 multiprotocol friendly. 

- Handling the signature difference of sd_softdevice_enable(...) across singleprotocol and multiprotocol softdevices
- Enabling Softdevice event handling for the concurrent protocol stack
  - Adding an optional semaphore pointer to AdafruitBlufruit together with a corresponding setter method
  - Extending SD_EVT_IRQHandler() to signal the additional semaphore, as well, if set

I believe these changes are 100% backward compatible.

NB: These changes would be needed for _any_ alternative protocol-stack running concurrently with BLE.


libraries/Bluefruit52Lib/src/bluefruit.cpp
```diff
@@ -676,6 +685,8 @@ void SD_EVT_IRQHandler(void)
   // Notify both BLE & SOC Task
   xSemaphoreGiveFromISR(Bluefruit._soc_event_sem, NULL);
   xSemaphoreGiveFromISR(Bluefruit._ble_event_sem, NULL);
+  // Notify parallel multiprotocol Task, if any
+  if (Bluefruit._mprot_event_sem!=NULL) xSemaphoreGiveFromISR(*Bluefruit._mprot_event_sem, NULL);
```
hathach, line 689: please wrap all the ant related code inside `#ifdef ANT_LICENSE_KEY` there must be zero code addition when ANT key is not defined.

libraries/Bluefruit52Lib/src/bluefruit.h
```diff
@@ -178,6 +178,15 @@ class AdafruitBluefruit
 
     BLEConnection* Connection(uint16_t conn_hdl);
 
+    /*------------------------------------------------------------------*
+     * Optional semaphore for additional event handlers for SD event.
+     * It can be used for handling non-BLE  SD events 
+     *------------------------------------------------------------------*/
+    void setMultiprotocolSemaphore(SemaphoreHandle_t* p_mprot_event_semaphore)
```
hathach, line 185: It is not clear to me why semaphore has to passed from application, we should create it  when enable the the stack just like the ble semaphore.

libraries/Bluefruit52Lib/src/bluefruit.cpp
orrmany: Ok, I´ll do that soon

libraries/Bluefruit52Lib/src/bluefruit.h
orrmany: Ok, I see no reason why that could not work. I need a day, or two until I can test that variant due to other projects, but I try to prioritize this PR.

orrmany: Oh, I forget the reason, but now I recall. @hathach: I need to be able to "take" that semaphore, which I can't do if that is a private member of the Bluefruit instance.  What do you propose? 

- Either the semaphore would have to be made public, 
- or it has to be made protected and my ANT-handler class needs to be made "friend" of Bluefruit, 
- or a getter method is needed for the memory address of the semaphore. 
Either of these solution is equivalent to the current proposal having a semaphore external to Bluefruit...

hathach: @orrmany In this case, since bluefruit.cpp has no other usage for the semaphore. I would suggest to just have an weak function hook() for mprot. Which is defined in your own application.

libraries/Bluefruit52Lib/src/bluefruit.cpp
```diff
@@ -676,6 +687,10 @@ void SD_EVT_IRQHandler(void)
   // Notify both BLE & SOC Task
   xSemaphoreGiveFromISR(Bluefruit._soc_event_sem, NULL);
   xSemaphoreGiveFromISR(Bluefruit._ble_event_sem, NULL);
+#ifdef ANT_LICENSE_KEY
+  // Notify parallel multiprotocol Task, if any
+  if (Bluefruit._mprot_event_sem)  xSemaphoreGiveFromISR(Bluefruit._mprot_event_sem, NULL);
```
hathach, line 692: Let's make this a weak function hook. The name should be something like `sd_evt_handler_hook()` that is defined by application. This way your library take a full control of Create and usage of the semaphore without sharing it to the current bluefruit.cpp.

orrmany:
" Let's make this a weak function hook. " I do not get it, sorry. Please
explain what do you mean?

Ha Thach <notifications@github.com> ezt írta (időpont: 2020. júl. 10., P,
14:19):

libraries/Bluefruit52Lib/src/bluefruit.cpp
orrmany: @hathach: Do you consider, that here the Bluefruit code is inside an interrupt handler? Calling a user-defined function from inside an interrupt handler does not sound as a wise SW-engineering practice for me. Too big risk is the application code will screw-up the efficiency/speed of the interrupt handler. "Giving" a semaphore is an action with a well-defined and well-limited impact on the interrupt processing speed, I think that is much more safe and robust for Adafruit than calling arbitrary user-land code via a function hook, IMHO...

hathach: Yeah, I think you are right on this. That is not really safe way to do so. I do overlook that.

libraries/Bluefruit52Lib/src/bluefruit.cpp
```diff
@@ -310,7 +313,15 @@ bool AdafruitBluefruit::begin(uint8_t prph_count, uint8_t central_count)
   #error Clock Source is not configured, define USE_LFXO or USE_LFRC according to your board in variant.h
 #endif
 
+  /*------------------------------------------------------------------
+  ** BLE only Softdevices have 2-args sd_softdevice_enable()
+  ** BLE & ANT+ Softdevices have 3-args sd_softdevice_enable()
+  */
+  #ifdef ANT_LICENSE_KEY
```
hathach, line 320: hmm, this isn't aligned well, but that is not really an issue to further delay this PR.

orrmany:
Hi @hathach, I've been away from embedded programming for more than year for various reasons, but now I'm back and I am trying to uplift my previous ANT+ hacks to the latest upstream release (v.9.3.0). I see, that there have been a major refactor here. What I seek your help with is this part from src/usb/uf2/uf2cfg.h:
Previously my code in src/usb/uf2/uf2cfg.h looked like this
> #ifdef S340 //S340 uses more flash
> #define USER_FLASH_START   0x31000
> #else
> #define USER_FLASH_START   0x26000
> #endif
> #define USER_FLASH_END     0xAD000 // Fat Fs start here
That is, if the S340 softdevice was chosen, then the flash start has been shifted with 0xB000 to make room for the bigger ANT+BLE softdevice. Now the current src/usb/uf2/uf2cfg.h in v.9.3.0 looks like this: 
> #define USER_FLASH_START          MBR_SIZE // skip MBR included in SD hex
> #define USER_FLASH_END            0xAD000
Do you have any recommendation where and how to adjust the flash-allocation in your UF2 code for the s340 softdevice? Thanks in advance

PR #519, opened 2020-06-25
Add truncate method to Adafruit_LittleFS_File.

pyro9:
Adds a truncate method with semantics like Unix ftruncate and a version with no parameter that truncates to the current file
position.

### DIFF
--- a/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
+++ b/libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
@@ -272,6 +272,32 @@ uint32_t File::size (void)
   return ret;
 }
 
+bool File::truncate (uint32_t pos)
+{
+  uint32_t ret=0;
+  _fs->_lockFS();
+  if (!this->_is_dir)
+  {
+    ret = lfs_file_truncate(_fs->_getFS(), _file, pos);
+  }
+  _fs->_unlockFS();
+  return ( ret == 0 );
+}
+
+bool File::truncate (void)
+{
+  uint32_t ret=0;
+  uint32_t pos;
+  _fs->_lockFS();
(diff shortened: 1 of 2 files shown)


libraries/Adafruit_LittleFS/src/Adafruit_LittleFS_File.cpp
```diff
@@ -272,6 +272,32 @@ uint32_t File::size (void)
   return ret;
 }
 
+bool File::truncate (uint32_t pos)
+{
+  uint32_t ret=0;
+  _fs->_lockFS();
+  if (!this->_is_dir)
```
hathach, line 279: this seems to return true when the FIle is directory, I am not sure about that semantic, should we return false with directory !!!

pyro9: Yes, upon thought, truncate on a directory should return false. Pull updated.

hathach: please use the `LFS_ERR_ISDIR` for error code with directory

pyro9: I'm not sure, change type to size_t and return error code or new size of file?
The other methods seem to avoid exposing LFS error codes and return 0 or false for inappropriate operations performed on a directory.

hathach: @pyro9 I am not mentioning the returned code, I mean changing default `uint32_t ret=1;` into `int ret=LFS_ERR_ISDIR`;

pyro9: Now I understand. I have made the change.

PR #532, opened 2020-07-13
label new/delete as weakly bound.  So that apps can redefine as needed

geeksville:
Sometimes app developers need to redefine new/delete for debugging or other purposes.  This changes the definitions to have the weak attribute, so that will be allowed by the linker rather than throwing an error.

### DIFF
--- a/cores/nRF5/new.cpp
+++ b/cores/nRF5/new.cpp
@@ -23,27 +23,27 @@
 #include <sys/stat.h>
 #include <malloc.h>
 
-void *operator new(size_t size) {
+void __attribute__((weak)) *operator new(size_t size) {
   return rtos_malloc(size);
 }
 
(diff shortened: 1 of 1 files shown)


hathach:
can you explain it in more detail

geeksville:
sure.  It is often super useful to redefine new/delete so you can debug, track all allocations.  And in most (all? it might be a requirement of the C++ language spec?) libraries new/delete are declared with weak binding, so applications can define their own implementation if needed.

The adafruit_nrf52 lib was not marking their implementation as weakly bound.  So any attempt to redefine it would be declared illegal by the linker.

hathach:
Closed since I don’t think it is necessary. For debug purpose please modify the func directly. We can’t have all function as weak for override

PR #534, opened 2020-07-13
Add board variant definitions for RAK815 nrf52832+lora+gps board

geeksville:
Hi,

I don't know if ya'll are interested in including support for new board types, but if you are here's the variant files for the rak815 board. It seems to me that platformio requires that all variants live in this directory (do you happen to know if this is true?).

If you aren't interested in including other board types in this project that's fine also. But I wanted to make these files available so that other projects besides mine could 'just work' with this board.

If you do want to support other boards, let me know and I can also send in a PR adding the definitions for the bootloader.

Thank you for your great project!

### DIFF
--- a/variants/rak815/variant.cpp
+++ b/variants/rak815/variant.cpp
@@ -1,0 +1,81 @@
+/*
+  Copyright (c) 2014-2015 Arduino LLC.  All right reserved.
+  Copyright (c) 2016 Sandeep Mistry All right reserved.
+  Copyright (c) 2018, Adafruit Industries (adafruit.com)
+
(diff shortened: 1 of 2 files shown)


hathach:
We are cool with adding 3rd party boards to the repo, as long as you are willing to provide support for issue with the board. Which mostly is us tagging you for issue on github and adafruit support forum. If you are committed to provide support, feel free to open PR.

For PIO variant indexing, you should ask them directly, since they use custom crawler / indexer.

hathach:
closed due to lack of user response.

Issue #514, opened 2020-05-27
Discovery failed on Manufacture Characteristic in Device Infomation Service.


arms22:
**Describe the bug**
Discovery failed on Manufacture Characteristic in Device Infomation Service.

**Set up (please complete the following information)**
  - Raytac MDBT50Q-RX Dongle
  - Windows 10
  - Serial debug log (enable IDE's Debug Mode Level to 1 or 2)  
```text
[GAP   ] Conn Interval= 20.00 ms, Latency = 0, Supervisor Timeout = 2000 ms
[BLE   ] BLE_GATTC_EVT_CHAR_VAL_BY_UUID_READ_RSP : Conn Handle = 0
[DISC  ] [SVC] Handle start = 1
[BLE   ] BLE_GATTS_EVT_EXCHANGE_MTU_REQUEST : Conn Handle = 0
[GAP   ] ATT MTU is changed to 131
[BLE   ] BLE_GATTC_EVT_PRIM_SRVC_DISC_RSP : Conn Handle = 0
[DISC  ] [SVC] Service Count: 1
[DISC  ] [SVC] Found 0x180A, Handle start = 14, end = 24    <-- Device Infomation Service
-----------------
[DISC  ] [CHR] Handle start = 14, end = 24
[BLE   ] BLE_GATTC_EVT_CHAR_DISC_RSP : Conn Handle = 0
[DISC  ] [CHR] Characteristic Count: 5
[DISC  ] [CHR] Found 0x2A24, handle = 16 
-----------------
[BLE   ] BLE_GATTC_EVT_READ_RSP : Conn Handle = 0
[DISC  ] [CHR] Handle start = 14, end = 24
[BLE   ] BLE_GATTC_EVT_CHAR_DISC_RSP : Conn Handle = 0
[DISC  ] [CHR] Characteristic Count: 5
[DISC  ] [CHR] Found 0x2A25, handle = 18
-----------------
[BLE   ] BLE_GATTC_EVT_READ_RSP : Conn Handle = 0
[DISC  ] [CHR] Handle start = 14, end = 24
[BLE   ] BLE_GATTC_EVT_CHAR_DISC_RSP : Conn Handle = 0
[DISC  ] [CHR] Characteristic Count: 5
[DISC  ] [CHR] Found 0x2A26, handle = 20
-----------------
[BLE   ] BLE_GATTC_EVT_READ_RSP : Conn Handle = 0
[...]

PR #535, opened 2020-07-14
Fix  BLEDiscovery::discoverCharacteristic when the central device returns more than 4 Characteristics in a discovery request

pyro9:
Recover gracefully when a BLE peripheral returns more than 4 characteristics in a reply to a discovery request by clamping the reply's characteristic count to match what's actually in the truncated reply.

Should fix issue #514.

### DIFF
--- a/libraries/Bluefruit52Lib/src/BLEDiscovery.cpp
+++ b/libraries/Bluefruit52Lib/src/BLEDiscovery.cpp
@@ -109,7 +109,8 @@ uint8_t BLEDiscovery::discoverCharacteristic(uint16_t conn_handle, BLEClientChar
   // We could found more characteristic than we looking for. Buffer must be large enough
   enum { MAX_DISC_CHARS = 4 };
 
-  uint16_t bufsize = sizeof(ble_gattc_evt_char_disc_rsp_t) + (MAX_DISC_CHARS-1)*sizeof(ble_gattc_char_t);
+  // -1 because the first ble_gattc_char_t is built in to ble_gattc_evt_char_disc_rsp_t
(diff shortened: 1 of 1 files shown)


libraries/Bluefruit52Lib/src/BLEDiscovery.cpp
```diff
@@ -129,6 +130,9 @@ uint8_t BLEDiscovery::discoverCharacteristic(uint16_t conn_handle, BLEClientChar
     // timeout or has no data (due to GATT Error)
     if ( bytecount <= 0 ) break;
 
+    if ( disc_chr->count > MAX_DISC_CHARS ) // if we truncated the response, adjust the count to match
```
hathach, line 133: can you have this simple if in the same line. I prefer either having single line if or with an full bracket one.

```
// if we truncated the response, adjust the count to match
if ( disc_chr->count > MAX_DISC_CHARS ) disc_chr->count = MAX_DISC_CHARS;
```

libraries/Bluefruit52Lib/src/BLEDiscovery.cpp
```diff
@@ -109,7 +109,8 @@ uint8_t BLEDiscovery::discoverCharacteristic(uint16_t conn_handle, BLEClientChar
   // We could found more characteristic than we looking for. Buffer must be large enough
   enum { MAX_DISC_CHARS = 4 };
```
hathach, line 110: maybe we should increase this to 8, discovery is expensive operation and takes air time. I wasn't too sure which is max each pass. Look like the sizeof(ble_gattc_char_t) is not that big.

PR #536, opened 2020-07-14
Added I2C pin configuration

thaanstad:
To configure the I2C pins call Wire.setPins(SDA_Pin, SCL_pin) prior to Wire.begin()

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -39,7 +39,7 @@ class TwoWire : public Stream
     void begin(uint8_t);
     void end();
     void setClock(uint32_t);
-	void setPins(uint8_t pinSDA, uint8_t pinSCL);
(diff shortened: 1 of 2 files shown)


thaanstad:
I am kind of new to working on github with pull requests etc.  It looked like my version of the repo failed when building the StandardFirmataBLE.ino.  I don't think this is related to the code I edited is it?

libraries/Wire/Wire.h
```diff
@@ -39,6 +39,7 @@ class TwoWire : public Stream
     void begin(uint8_t);
     void end();
     void setClock(uint32_t);
+	void setPins(uint8_t pin_SDA, uint8_t pin_SCL);
```
hathach, line 42: please fix the indent (better to use spaces over tab), also the variable name should be pinSDA/pinSCL to be consistent with constructor.

PR #554, opened 2020-08-12
Power fixes

Nenik:
@hathach this should be the right set of changes, let's see.
(sorry for the noise, not familiar with GH pull requests)


hathach:
@Nenik no problem at all, everyone has his first time :). Btw, I am afraid I have to ask you to re-do the modification but instead of `develop`, please use the `master` branch. As rule of thumb, all the the PR should target the master, `develop` is used as internal branch, (currently it is implementing the LESC feature making use of CryptoCell). 

I think you could just copy the port_cmsis_systick.c to another location, then check out to master, and copy the file back (since it has minimal changes so far). Then push force and change the target branch from `develop` to `master`. There may be other faster way as well (I am not too good with git on this as well). Let's me know if you have any issues doing this.

Nenik:
Closing in favor of #555

PR #556, opened 2020-08-13
Move CMSIS headers to work with PIO platform

KK1423:
Moved files to location expected by Platformio [build script](https://github.com/platformio/builder-framework-arduino-nrf5/blob/d7817f50a6548025711a5e1a911987d1b072a9b4/adafruit.py#L107).


hathach:
CMSIS of this repo is updated to v5 since 0.20.0. Which has directory org changed into `Core`. PIO build script should be updated to reflect the latest changes.

ladyada:
@ivankravets ^ see possible PIO bug report for upgraded CMSIS (we've updated both nRF and SAMD to 5 for tensorflow support and basic improvementness)

ivankravets:
@KK1423  do you use the latest stable PlaformIO dev-platform? Everythin should work without any issues.

KK1423:
I'm using the platform with my fork of this git repo, rather than the package it downloads by default (likely version <= 19). Perhaps the build script should decide the include dir based on the version.

ivankravets:
If you need the latest version of this core, please use Arduino IDE. This is the only official IDE solution for Adafruit products, see https://learn.adafruit.com/adafruit-arduino-ide-setup/arduino-1-dot-6-x-ide

You can also file a feature request at https://github.com/platformio/platform-nordicnrf52/issues to update PlatformIO's official support for this core to the latest version. However, we don't have any ETA when we can update this. We work on OSS feature requests based on their priority and community votes.

> Perhaps the build script should decide the include dir based on the version.

We created the ecosystem where any silicon vendor or hardware manufacturer can create own dev-platform and maintain it. You can check how other companies do this, for example, Espressif:
- https://github.com/espressif/arduino-esp32/blob/master/tools/platformio-build.py
- https://github.com/esp8266/Arduino/blob/master/tools/platformio-build.py

So, their customers can use PlatformIO even with the development version of their Arduino core.

PR #557, opened 2020-08-14
Added resumeLoop functionality

kevinfrei:
In order to pause loop functionality, suspendLoop works great, but without resumeLoop (and resumeLoopFromISR), suspendLoop is effectively "terminateLoop" as there's no way to start it back up.


kevinfrei:
Oops. Will re-submit in a minute...

PR #558, opened 2020-08-14
Add resumeLoop functionality

kevinfrei:
Without `resumeLoop`/`resumeLoopFromISR` functions, suspendLoop is the same as terminateLoop. This enables an interrupt to trigger looping (using `resumeLoopFromISR`), which can then operate for some time, then suspend itself.

Example use: Use an interrupt to detect when a switch in a key matrix is pressed. Debouncing with only interrupts is nearly impossible. This allows the interrupt to trigger a normal "keymatrix scan & debounce" loop, which runs normally until debouncing time is over, then the loop can be suspended (thereby saving tremendous amounts of power).

### DIFF
--- a/cores/nRF5/main.cpp
+++ b/cores/nRF5/main.cpp
@@ -102,11 +102,14 @@ void suspendLoop(void)
 
 void resumeLoop(void)
 {
-  vTaskResume(_loopHandle);
-}
-
-void resumeLoopFromISR(void) {
-  xTaskResumeFromISR(_loopHandle);
+  if ( isInISR() ) 
+  {
+    xTaskResumeFromISR(_loopHandle);
+  } 
+  else
+  {
+    vTaskResume(_loopHandle);
+  }
 }
 
 extern "C"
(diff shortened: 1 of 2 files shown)


cores/nRF5/main.cpp
```diff
@@ -100,6 +100,15 @@ void suspendLoop(void)
   vTaskSuspend(_loopHandle);
 }
 
+void resumeLoop(void)
+{
+  vTaskResume(_loopHandle);
+}
+
+void resumeLoopFromISR(void) {
+  xTaskResumeFromISR(_loopHandle);
```
hathach, line 109: There is no need to for FromISR version, you could detect the ISR context and call the API using `isInISR()` as this 
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/utility/SoftwareTimer.cpp#L70

kevinfrei:
Updated the branch with the suggestion. Works like a charm (I hadn't seen that API while digging around in FreeRTOS)

hathach:
> Updated the branch with the suggestion. Works like a charm (I hadn't seen that API while digging around in FreeRTOS)

it is specific to ARM cortex M3/M4/M7 only, will merge after ci passed.

cores/nRF5/main.cpp
kevinfrei: Done!

PR #560, opened 2020-08-24
Minor cleanup

henrygab:
Leftovers from #496.
* [x] Remove semaphore from `tone()`
* [x] Remove 'inline' ... it's not related in any way to compiler optimization, but rather defines whether the definition can be within a header file, which is not applicable here  (it's not `__attribute__((always_inline))`)

### DIFF
--- a/cores/nRF5/Tone.cpp
+++ b/cores/nRF5/Tone.cpp
@@ -64,7 +64,6 @@ class TonePwmConfig {
         uint8_t nrf_pin;              //< the nrf pin for playback
         nrf_pwm_task_t task_to_start; //< Whether to start playback at SEQ0 or SEQ1
         nrf_pwm_short_mask_t shorts;  //< shortcuts to enable
-
     public:
         bool ensurePwmPeripheralOwnership(void);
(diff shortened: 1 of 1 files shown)


cores/nRF5/Tone.cpp
```diff
@@ -74,7 +73,7 @@ class TonePwmConfig {
 };
 TonePwmConfig _pwm_config;
 
-inline static bool _is_pwm_enabled(NRF_PWM_Type const * pwm_instance) {
+static bool _is_pwm_enabled(NRF_PWM_Type const * pwm_instance) {
```
hathach, line 76: inline is still OK, it is more like suggestion to gcc. It is up to gcc to decide e.g Ofast may inline the function, but Os may not since it increase the size. attribute always inline is more like directive to gcc to force it to do so. So yeah, it is normally not an issue on non timing-sensitive function. Anyway, I normally just put inline where I would like, and leave the decision to gcc.

hathach (approved):
Thanks for removing the lamda function, I think I am too conservative for it. Maybe I should update my C/C++ knowledge, still writing C99 for now.

PR #561, opened 2020-08-24
[WIP] prototype fix for mixing analogWrite() and digitalWrite()

### LINKED ISSUE #447: digitalWrite() doesn't override analogWrite()
**Describe the bug**
Once PWM is started on a pin with `analogWrite()`, the pin can't be set with `digitalWrite()` anymore.

**Set up**
NRF52840

**To Reproduce**
```
analogWrite(LED_GREEN,128);
digitalWrite(LED_GREEN,LOW);
```

**Expected behavior**
PWM on `LED_GREEN` goes to 50%, then pin goes low.
**Actual behavior**
-> pin stays at 50%

henrygab:
Fix #447

### DIFF
--- a/cores/nRF5/wiring_digital.c
+++ b/cores/nRF5/wiring_digital.c
@@ -25,15 +25,17 @@
 extern "C" {
 #endif
 
+static bool removePinFromAllHwPWMInstances(uint8_t pin); // definition in HardwarePWM.cpp
+uint8_t LastUsedNrfPinAsAnalog[NUMBER_OF_PINS];          // extern access from HardwarePWM.cpp
+
 void pinMode( uint32_t ulPin, uint32_t ulMode )
 {
   if (ulPin >= PINS_COUNT) {
     return;
   }
 
-  ulPin = g_ADigitalPinMap[ulPin];
-
-  NRF_GPIO_Type * port = nrf_gpio_pin_port_decode(&ulPin);
+  uint32_t nrfPin = g_ADigitalPinMap[ulPin];
+  NRF_GPIO_Type * port = nrf_gpio_pin_port_decode(&nrfPin);
 
   // Set pin mode according to chapter '22.6.3 I/O Pin Configuration'
   switch ( ulMode )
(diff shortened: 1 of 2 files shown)


cores/nRF5/wiring_digital.c
```diff
@@ -98,13 +100,26 @@ void pinMode( uint32_t ulPin, uint32_t ulMode )
...
   }
 
-  ulPin = g_ADigitalPinMap[ulPin];
+  uint32_t nrfPin = g_ADigitalPinMap[ulPin];
+
+  // Caller is mixing analogWrite() and digitalWrite()
+  // Arduino API supports this arbitrary mixing.
+  // Use __builtin_expect() to request compiler
+  // to optimize for when this is false
+  if (__builtin_expect(!!(LastUsedNrfPinAsAnalog[nrfPin]), 0)) {
```
hathach, line 116: I don't think `__builtin_expect()` can help with anything here, It is like we "hope compiler could do the best for us". I would rather to remove it for a clean code unless you could point out the difference between generated codes.

hathach: also we could just use a loop() to go through the HWPWM list here to check for the pins. Less work than the `LastUsedNrfPinAsAnalog[NUMBER_OF_PINS]`. We will optimize later when people complain about the performance if any.

henrygab: @hathach ...  You want to loop through 12 / 16 PWM module's configured pins, for **_each and every_** digitalWrite() call?  This means every time any pin is toggled....  at least 64 reads of the PWM module registers across the AHB bus (@16MHz ... not 64MHz) ... that's not a change I would support.

I believe it is much better to only attempt to remove when the sketch requires it.  In fact, it was my understanding that `digitalWrite()` is considered critical-path, which is why I've optimized to minimize impact on `digitalWrite()`.  The goal is to minimize impact on any high-speed toggling of the pins (e.g., bitbang versions of various protocols).

Since this is a simple change (I think it's basically ready as-is, pending testing), and I am much less comfortable blindly attempting to remove a pin from the PWM modules at **_every_** call to `digitalWrite()`, I am going to keep this PR using this method (unless an even better solution pops up).

Of course, please indicate if you have coded a different solution, as I would then be thrilled to close this PR!

hathach: I really don't like to have digital/analog to set/get an array upon call, but I think you are persistent. Let's make it this way
- use bitmask uint64_t instead of array
- bitmask is static member of HardwarePWM 
- bitmask is Aduino Pin (not native nRF)

These will make the code cleaner I think

hathach: I still like using pinMode(Output) to remove analogWrite() less code to worry about. I think we should do that first and having this pending for sometime until people start complaining

henrygab:
@hathach -- I will close this PR then.

hathach:
> @hathach -- I will close this PR then.

Thanks @henrygab , we will see if people complaining about this and fix later on :smile:

PR #562, opened 2020-08-25
clean up pwn and analog

hathach:
- replace built-in atomic with freeRTOS critical functions
- mostly clean up and (re)move thing around

It may take a couple more follow-up PR, but  I will switch to other task. Therefore this one is good for a review.


cores/nRF5/HardwarePWM.cpp
```diff
@@ -350,3 +270,58 @@ uint8_t HardwarePWM::freeChannelCount(void) const
...
+    return false;
+  }
+
+  if ( this->_owner_token != 0 ) return false;
+  if ( this->usedChannelCount() != 0 ) return false;
+  if ( this->enabled() ) return false;
+
+  // This function must not be called within ISR
+  taskENTER_CRITICAL();
+  _owner_token = token;
```
henrygab, line 292: This is not correct ... you have introduced a bug.
The old code:

```C++
uintptr_t newValue = 0U;	
return this->_owner_token.compare_exchange_strong(newValue, token);
```

In contrast, the code inside the critical section here is modifying the value unconditionally ... a clear race condition.

cores/nRF5/Tone.cpp
```diff
@@ -332,14 +332,8 @@ bool TonePwmConfig::applyConfiguration(uint32_t pin) {
     nrf_pwm_shorts_set(_PWMInstance, this->shorts);
     nrf_pwm_int_set(_PWMInstance, 0);
 
-    // static sequence value ... This is the value actually used
-    // during playback ... do NOT modify without semaphore *and*
-    // having ensured playback has stopped!
-    static uint16_t seq_values; // static value
```
henrygab, line 342: Do you realize that, by this change, you must add synchronization to `this->duty_with_polarity`? (or otherwise ensure not playing before changing this).

This makes the code more _fragile_; I recommend against this change.

cores/nRF5/HardwarePWM.cpp
```diff
@@ -350,3 +278,72 @@ uint8_t HardwarePWM::freeChannelCount(void) const
...
+
+  if (this->enabled()) {
+    LOG_LV1("HwPWM", "attempt to release ownership when PWM peripheral is still enabled");
+    return false; // if it's enabled, do not allow ownership to be released, even with no pins in use
+  }
+
+  if (isInISR())
+  {
+    UBaseType_t intr_status = taskENTER_CRITICAL_FROM_ISR();
+    _owner_token = 0;
```
henrygab: Bug -- you added a race condition by unconditionally changing the value.

cores/nRF5/HardwarePWM.cpp
```diff
@@ -350,3 +278,72 @@ uint8_t HardwarePWM::freeChannelCount(void) const
...
+    taskEXIT_CRITICAL_FROM_ISR(intr_status);
+  }else
+  {
+    taskENTER_CRITICAL();
+    _owner_token = 0;
```
henrygab: Bug -- you added a race condition...

cores/nRF5/Tone.cpp
```diff
@@ -370,13 +370,9 @@ bool TonePwmConfig::startPlayback(void) {
...
 bool TonePwmConfig::stopPlayback(bool releaseOwnership) {
 
-    bool notInIsr = !isInISR();
-
     if (!_HwPWM->isOwner(TonePwmConfig::toneToken)) {
-        if (notInIsr) {
-            LOG_LV2("TON", "Attempt to set noTone when not the owner of the PWM peripheral.  Ignoring call....");
-        }
-        return false;
+      LOG_LV2("TON", "Attempt to set noTone when not the owner of the PWM peripheral.  Ignoring call....");
```
henrygab: If `noTone()` called from ISR, and compiled at debug level 2, will this hang if the serial buffer is full?

libraries/Bluefruit52Lib/examples/Hardware/tone_happy_birthday/tone_happy_birthday.ino
```diff
@@ -0,0 +1,146 @@
+// Circuit Playground Birthday Candles
```
henrygab, line 1: 🎉 🎈 🍰 👍

libraries/Servo/src/nrf52/Servo.cpp
```diff
@@ -47,53 +47,58 @@ uint8_t Servo::attach(int pin, int min, int max)
...
-  pinMode(pin, OUTPUT);                                   // set servo pin to output
-  servos[this->servoIndex].Pin.nbr = pin;
 
-  if (min < MIN_PULSE_WIDTH) {
-    min = MIN_PULSE_WIDTH;
-  }
-  if (max > MAX_PULSE_WIDTH) {
-    max = MAX_PULSE_WIDTH;
-  }
+  if (min < MIN_PULSE_WIDTH) min = MIN_PULSE_WIDTH;
```
henrygab, line 51: The lack of braces is a common cause of bugs.  therefore, I recommend against removing them.   Even if you place them on the same line, I recommend keeping the curly braces.

cores/nRF5/HardwarePWM.cpp
hathach: I don't see anything except SoftDevice could do a race condition here. SD is clearly won't run be able to do a race condition here.

cores/nRF5/Tone.cpp
hathach: I don't see any issue with seq got changed. it only happens when use call tone() again, it is OK to got value changed.

cores/nRF5/HardwarePWM.cpp
hathach: The condition won't happen, nothing go through the task critical

cores/nRF5/Tone.cpp
hathach: user should not call it withint ISR()

libraries/Servo/src/nrf52/Servo.cpp
hathach: OK, this actually make sense, but it is not eye candy. Though I will change it.

hathach:
Thank you very much for your review,
 
> 1. I **_strongly_** recommend **_against_** removal of the atomic operation.  The existing code works, while yours changes add bugs and make the code more complex.

There should be no race condition, the mutex is ensured by critical API(). There is nothing could take the control of CPU between Enter/Exit critical except for BLE softdevice since it has highest priority, higher than kernel's one. 

> 2. If the serial port buffer is full, and `noTone()` is called from an ISR, what result? (will it hang waiting for space in serial output buffer?)
> 
> I have attempted to comment on each change that appeared to cause bugs.

User must not call tone() or noTone() within ISR. Like my current BLE API, anything API got called within ISR will mostly like to hanged. There is way to defer function to thread-context.

cores/nRF5/HardwarePWM.cpp
henrygab: #### _Yes, there is a race condition._

Your code is not equivalent.  Read the definition of `compare_exchange_strong()` to understand.

#### Example race condition

1. Start with `_owner_token == nullptr`
2. Task 1: (low priority task) runs to line 290 quoted above
3. ISR: interrupt fires, pre-empting Task 1
4. ISR: calls the function with owner token 0xFEED, which **_succeeds_**
    * `_owner_token == 0xFEED`
    * Note: ISR could also have caused a higher-priority task to execute, which called the function...
5. Task 1: Resumes operation at line 290
6. Task 1: sets _owner_token to 0xBAADF00D (overwrites the existing owner 0xFEED)
    * `_owner_token == 0xBAADF00D`
7. Task 1: returns an indication of **_success_**

Two tasks believe they successfully obtained ownership, and then corrupt device state.

cores/nRF5/Tone.cpp
henrygab: If you cannot see the problem, I give up.

cores/nRF5/HardwarePWM.cpp
henrygab: It's not the task critical.  It's the unconditional setting of the value.

You clearly did not understand the code you were replacing.

cores/nRF5/HardwarePWM.cpp
henrygab: Same as prior comment.

cores/nRF5/Tone.cpp
henrygab: Really?  Don't you recall the old code, which called `noTone()` from ISR to stop itself when counter was zero?

Clearly, if BSP code did this, callers could do so also.

You are changing the contract, and adding a new restriction to sketches.

**_Don't do this_**.  You will pay for this later with very difficult to track down bugs from user sketches.

henrygab:
> There should be no race condition, the mutex is ensured by critical API(). There is nothing could take the control of CPU between Enter/Exit critical except for BLE softdevice since it has highest priority, higher than kernel's one.

Yes, there *is* a race condition.  I give details above.  READ IT.


> User must not call tone() or noTone() within ISR. Like my current BLE API, anything API got called within ISR will mostly like to hanged. There is way to defer function to thread-context.

The BSP itself called `noTone()` from its own ISR.  Therefore, user sketches were allowed to do so.   In other words, your own code is what made it OK for sketches to call `noTone()` from ISR.  You will now break stuff.

#### Question

Why are you changing **_working, bug-free code_**, after telling me so many times that you only make changes when 
someone complains of a bug?

#### Summary

* [ ] I **strongly** recommend **_against_** the removal of the atomic operation.
* [ ] I recommend **_against_** making it illegal to call `noTone()` from ISR.

I cannot spend hours trying to teach you anymore.  I have tried my best to warn you away.  If you resolve the issues because you cannot see the issue, I will **_not_** stop you anymore.

cores/nRF5/HardwarePWM.cpp
hathach: that is mostly true but only without critical section. https://www.freertos.org/taskENTER_CRITICAL_taskEXIT_CRITICAL.html 
there is no ISR that can interrupt the task once the API is invoked.

cores/nRF5/Tone.cpp
hathach: That is not that bad, the Bluetooth API is used from the start, and there should is no problem so far. User shouldn't do much processing inside any ISR(). 

PS: The previous implementation is copied from other repo, which I am too lazy to change. Now I don't see any reason to allow the API to be called in ISR.

hathach:
@henrygab thanks again for more review, please check out the taskENTER_CRITICAL() to see how it provide the mutex, the C++ atomic in the other way may not provide the same level of race condition guarding in efficient way. As mentioned multiple time before I suspect it uses cspie which will also disable the Softdevice task that handles timing-sensitive radio event. I would be happy if you could tell me more on the implementation of the C++ atomic.

cores/nRF5/HardwarePWM.cpp
hathach: Ah I see the racing scenario that you mentioned, we can move the code a bit to fix the racing.

hathach:
> Why are you changing working, bug-free code, after telling me so many times that you only make changes when
someone complains of a bug?

Yeah, I only want to add code when people is complaining, though I would like to maintain an clean code that I could easily read and understand as well.

cores/nRF5/Tone.cpp
hathach: AFAIK initializeFromPulseCountAndTimePeriod() is the only function that change the `duty_with_polarity`, And the playback is stopped before it is called. There should be no conflict when value changed when playing back, or I am missing something  !!

cores/nRF5/Tone.cpp
henrygab: You are adding a new restriction ... a breaking change to the API.  I recommend against your change.

You do whatever you want.

cores/nRF5/HardwarePWM.cpp
```diff
@@ -289,10 +289,13 @@ bool HardwarePWM::takeOwnership(uintptr_t token)
...
   taskENTER_CRITICAL();
-  _owner_token = token;
+  if ( this->_owner_token == 0 )
+  {
+    _owner_token = token;
+  }
   taskEXIT_CRITICAL();
 
-  return true;
+  return _owner_token == token;
```
henrygab, line 298: You have removed the 'atomic' for `_owner_token`.
You have not marked it volatile.
I have no idea if checking `_owner_token` outside the critical section is "safe" with your code.
You do this at your own risk.

hathach:
> I still strongly recommend against removal of the use of the atomic.
> You say you don't understand it's use.... but you can use a search engine.
> 
> "raymond chen compare_exchange_strong"
> 
> Regardless, it's cleaner because it's OS-independent. It never disables interrupts, and it has no effect on the softdevice. Your code does.

Ah thanks, I just google to that article here https://devblogs.microsoft.com/oldnewthing/20180329-00/?p=98375 . This discussion also explain how it is interrupt-safe as well https://community.arm.com/developer/ip-products/processors/f/cortex-m-forum/10361/ldrex-strex-on-the-m3-m4-m7 . It looks like the atomic implementation uses the LDREX/STREX on ARM cortex which is indeed 100% fit for this ownership purpose. You should really point me to the article earlier, I am so skeptical of the implementation, I should have done some research instead of just simply asking though (in my defense it is hard to get over the laziness :smile: ). Thank you for your passion, I have learned something new today.

hathach:
@henrygab I am persuaded by your persistence with the  takeOwnership/releaseOwnership regarding the ISR. They should really not be called within ISR, any user that write that should really need to rewrite the sketch. Normally I would just either ignore that or just simply drop the LOG message. However, since these API are new and there are potentially issue with the implementation, having the debug log is indeed useful. Thank you for your review so far.

hathach:
thank you very much for the in-depth review, I am glad that you are pushing me to learn new thing. I will be using the C11 atomic for cortex M mcu from now when possible. On other platform there maybe no equivalent of instruction, therefore an appropriate research should be done for each platform. But it is hard to find an MCU that is not ARM nowsaday.

dhalbert:
Sadly these are not available on the M0.

henrygab:
Really?  I thought that they were defined for M0 as something similar to:

<details><summary>click to expand / collapse code</summary><P/>

```C
int __atomic_fetch_add_4(int *d, int val, int mem)
{
	uint32_t primask = __get_PRIMASK();
	__disable_irq();
	*d += val;
	__set_PRIMASK(primask);

	return *d;
}
int __atomic_fetch_sub_4(int *d, int val, int mem)
{
	uint32_t primask = __get_PRIMASK();
	__disable_irq();
	*d -= val;
	__set_PRIMASK(primask);

	return *d;
}
```

</details>

Granted, this disables interrupts for a few instructions ... but are there synchronization primitives that don't disable interrupts on M0?

Or, were you saying that the `LDREX`/`STREX` instructions were not available on M0?

If the lack of those instructions is a concern, then unless there's the wierdness that nrf52 imposes via the softdevice, would the above read-modify-write with interrupts disabled be problematic on an M0?  (_obviously, the question is in comparison with a mutex or other synchronization primitive, each of which would appear to need to ensure an atomic compare-and-swap...._)

dhalbert:
> Or, were you saying that the `LDREX`/`STREX` instructions were not available on M0?

That's what's missing.

The SoftDevice has its own primitives, which don't block interrupts levels used by the SoftDevice, but block all other levels. It also provides mutexes.

```c
__STATIC_INLINE uint32_t sd_nvic_critical_region_enter(uint8_t * p_is_nested_critical_region)
{
  int was_masked = __sd_nvic_irq_disable();
  if (!nrf_nvic_state.__cr_flag)
  {
    nrf_nvic_state.__cr_flag = 1;
    nrf_nvic_state.__irq_masks[0] = ( NVIC->ICER[0] & __NRF_NVIC_APP_IRQS_0 );
    NVIC->ICER[0] = __NRF_NVIC_APP_IRQS_0;
    nrf_nvic_state.__irq_masks[1] = ( NVIC->ICER[1] & __NRF_NVIC_APP_IRQS_1 );
    NVIC->ICER[1] = __NRF_NVIC_APP_IRQS_1;
    *p_is_nested_critical_region = 0;
  }
  else
  {
    *p_is_nested_critical_region = 1;
  }
  if (!was_masked)
  {
    __sd_nvic_irq_enable();
  }
  return NRF_SUCCESS;
}

__STATIC_INLINE uint32_t sd_nvic_critical_region_exit(uint8_t is_nested_critical_region)
{
  if (nrf_nvic_state.__cr_flag && (is_nested_critical_region == 0))
  {
    int was_masked = __sd_nvic_irq_disable();
    NVIC->ISER[0] = nrf_nvic_state.__irq_masks[0];
    NVIC->ISER[1] = nrf_nvic_state.__irq_masks[1];
    nrf_nvic_state.__cr_flag = 0;
    if (!was_masked)
    {
      __sd_nvic_irq_enable();
    }
  }

  return NRF_SUCCESS;
}
```

hathach:
The sd_nvic_critical_region_enter/sd_nvic_critical_region_exit is equivalent to freeRTOS crtiical enter and exit, it block interrupt except one used by Softdevice and pause the scheduler. Softdevice interrupt should be never blocked since yeah, it will cause ble disconnection in most case. I am still amazed with  LDREX/STREX instruction on M3/M4 after working with the mcu for several years :), great feature.

Issue #497, opened 2020-05-05
[Feature] Migrate Interrupt handling to nrfx


czuvich:
I was trying to use the nrfx_gpiote functionality, and I was getting a compiler error saying that "GPIOTE_IRQHandler was already defined".  I dug into WInterrupts.c, and I noticed it is using the legacy drivers to configure interrupts.  nrfx_gpiote.h also defines the GPIOTE_IRQHandler as well and they use the same port.

It'd be really nice to use the nrfx library so that others can use more advanced features in the nrfx_gpiote stack.  In my use case, I'm performing pulse counting via PPI, Timer, and GPIOTE.  I am probably going to have to hack the WInterrupts.h to expose a way for me to get an available channel in the handler and use the legacy driver specifically for GPIOTE.

Now... I'm not an nRF52840 expert, so maybe I'm wrong about this... but please correct me if I am.  Thank you for the consideration.

henrygab:
@czuvich -- I think you are correct, but there may be a workaround.

### GPIOTE interrupt handler already defined

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/b03d6e5d814b48bcbaac39e38cb5f40e2b8e729f/cores/nRF5/WInterrupts.c#L180

The above line defines the implementation for the GPIOTE interrupt handler.  As there can be only one interrupt handler, it prevents the use of other GPIOTE interrupt handlers.

### Other bugs you'll likely run into

With the current release (0.21.0), `attachInterrupt()` would blindly use GPIOTE channels, even if you had previously configured them for other purposes.  This was fixed in #485 nine days ago (seven days before your opened this issue), but it's not made it into a release yet.  This fix will **_at least_** prevent the BSP from overwriting your configuration.....

### Potential workaround?

This workaround adds a eight-cycle delay (AHB runs at 16MHz vs. the CPU 64Mhz) due to the inter-peripheral signaling required.  This is less than the SoftDevice adds, so that may be OK for your use.

The nRF52840 has a peripheral called "EGU", for event generator unit.  (_EGU1 is reserved for use by SoftDevice._)  As an example, it appears that the following could work to get you another interrupt handler:

* Define an interrupt handler for EGU2 (SWI2_EGU2_IRQHandler) in your code
* Configure a PPI channel:  GPIOTE event --> EGU2 TASKS_TRIGGER[N] registers

This *should* give you an ISR that you own, triggered by the GPIOTE event (just a few cycles later).

[...]

PR #572, opened 2020-09-08
Add back missing nrfx source files

eugeneanikin:
As discussed in issue #497 this pr adds missing nrfx source files based on nrfx v2.1.0. To use these, one needs to add appropriate #defines in nrfx_config.h.

### DIFF
--- a/libraries/Bluefruit52Lib/src/BLECharacteristic.cpp
+++ b/libraries/Bluefruit52Lib/src/BLECharacteristic.cpp
@@ -724,7 +724,7 @@ bool BLECharacteristic::notify32(uint16_t conn_hdl, uint32_t num)
 
 bool BLECharacteristic::notify32(uint16_t conn_hdl, int num)
 {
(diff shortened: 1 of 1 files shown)


hathach:
many functions of these driver files are provided by Arduino API() e.g nrfx_spim/nrfx_twi, adding them all together may confuse user, and increase compiling time. Some won't even compile when enabled due to the conflict of IRQHandler(). How about only adding files that you/user need for now first. And we can work up gradually from that. I strip the nrfx also for this reason as well.

eugeneanikin:
Sure, I only need timer and pwm.

eugeneanikin:
Actually, the more I think about it, what I found confusing originally was that there were nrfx headers but no source files. So from that point of view, maybe something should be done about the unusable headers?

hathach:
Yeah, we should also remove unused header as well. I am just too lazy with maintenance work. I will be appreciate if you could remove those that is not used as well.

PR #574, opened 2020-09-09
Add missing nrfx source files

eugeneanikin:
As discussed in issue #497 this pr adds missing nrfx source files based on nrfx v2.1.0. To use these, one needs to add appropriate #defines in nrfx_config.h.
This version only includes the files I need.


hathach (approved):
it looks good, thank you for the PR

cores/nRF5/nordic/nrfx_config.h
```diff
@@ -14,6 +14,18 @@
 #define NRFX_SPIM1_ENABLED           0 // used as I2C
 #define NRFX_SPIM2_ENABLED           1
 
+#define NRFX_PWM_ENABLED 0
+#define NRFX_PWM0_ENABLED 0
+#define NRFX_PWM1_ENABLED 0
+#define NRFX_PWM2_ENABLED 0
+#define NRFX_PWM3_ENABLED 0
```
henrygab, line 21: If I understand correctly, the nRF52832 only has three PWM peripherals (`PWM0 .. PWM2`).  Should the `NRFX_PWM3_ENABLED` define be wrapped in `#ifdef` similar to `NRFX_SPIM3_ENABLED` is below?

hathach: hi @henrygab, no time no see. Indeed, 833 only has 3 pwm modules, however we should still defined the PWM3_ENABLED as 0 due to how the nrfx driver is implemented by Nordic. It checks for NRFX_PWM3_ENABLED valued without ifndef
https://github.com/NordicSemiconductor/nrfx/blob/master/drivers/src/nrfx_pwm.c#L39

henrygab: Hi @hathach , long time indeed.  Thank you for checking, and glad it won't use another PR.  👍

PR #586, opened 2020-10-04
fixed a bug that saved cccd data to a different offset after the second time.

ogatatsu:


hathach (approved):
superb ! thank you very much for spotting the issue.

PR #587, opened 2020-10-04
fix to not use mutex in ISR

ogatatsu:
The documentation states that mutexes must not be used within the ISR, so I fixed that.
https://www.freertos.org/a00124.html


hathach (changes requested):
Thank for spotting this, in this case, I think we better to just covert the current _mutex to _sem and use it as binary semaphore instead of introducing another sem.

cores/nRF5/Uart.cpp
```diff
@@ -142,6 +144,7 @@ void Uart::begin(unsigned long baudrate, uint16_t config)
   NVIC_EnableIRQ(IRQn);
 
   _mutex = xSemaphoreCreateMutex();
```
hathach, line 146: I think we can just use _end_tx_sem and remove the _mutex completely.

hathach (approved):
look great, thanks for your PR

PR #592, opened 2020-10-10
Add Pitaya Go support

windoze:
Add Pitaya Go from MakerDiary.com.
The bootloader was built from the source code at https://github.com/adafruit/Adafruit_nRF52_Bootloader, which already supports this board.


ladyada:
hi, pitaya can fork and create their own board support package - we dont manufacture it so we wouldn't be able to add support. thanks :)

PR #596, opened 2020-10-23
WIP adding BlueMicro 52840 to boards

jpconstantineau:
Preparing a pull request for your comments and suggestions.  
I still need to fully test it out within the IDE and with the bootloader (also PR'ed).

VID and PID have been handed out by openmoko in this [PR](https://github.com/openmoko/openmoko-usb-oui/pull/20).

We will need to add the bootloader binaries once the PR on the bootloader repo is approved.

Let me know if there is any specific requirements on "third party boards" you may have for pulling them into this repo.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -550,3 +550,53 @@ mdbt50qrx.menu.debug.l2.build.debug_flags=-DCFG_DEBUG=2
 mdbt50qrx.menu.debug.l3=Level 3 (Segger SystemView)
 mdbt50qrx.menu.debug.l3.build.debug_flags=-DCFG_DEBUG=3
 mdbt50qrx.menu.debug.l3.build.sysview_flags=-DCFG_SYSVIEW=1
+
+# ----------------------------------
+# BlueMicro nRF52840
+# ----------------------------------
+bluemicro52840.name=BlueMicro nRF52840
+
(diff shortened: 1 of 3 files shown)


hathach:
can fork and create their own board support package - we dont manufacture it so we wouldn't be able to add support. thanks :)

jpconstantineau:
Thanks for the comments.
Do you have any recommendations on how to manage a fork and keep in sync with the main BSP?

hathach:
it is up to you for when you think it is to do the sync, good idea to have it sync up on release.

jpconstantineau:
Sync on release is what I had in mind but it wasn't quite what I was really thinking about.   We don't really want to have 2 BSPs with adafruit boards, one official, the other forked with more boards in.  Is there a way to have a "Community BSP" that extends the main BSP?  This way, there is a clear separation while minimizing duplication.  If you know of documentation or a set of BSP that implements this, let me know...

ladyada:
hihi we can't have a BSP from adafruit with boards we dont own or test - we have tried it before and it just ends up with everyone frustrated. its pretty common for folks to fork our BSP and add their boards, here's one that forks our samd51 core
https://learn.sparkfun.com/tutorials/samd51-thing-plus-hookup-guide/setting-up-the-arduino-ide

PR #599, opened 2020-10-28
Add support for nrf52833 (including bluetooth) and softdevice S113 7.2.0

geeksville:
The older 6.1.1 soft device doesn't fully support the newer nrf52833 CPUs - in particular bluetooth needs 7.0 or later.  These changes add support for the 833 CPUs and newer soft devices.  I also added optional use of the S113 soft-device which saves a sizable amount of flash (which is especially useful on the low flash 832 and 833 devices).

I'll also send in a related PR for the bootloader to add support for 7.2 series soft devices 

(I'm making this draft PR so that I remember to clean things up and send in a final PR after my project slows down a little)


hathach:
@geeksville Thank you very much for your PR, currently we will only support 840 with version SD v6, since v7 has different flash layout which force existing user to upgrade bootloader to work with, also it need changes in other project such as circuitpython as well. However, I am open for some ground work like adding header, linker etc.. so that you could have minimized boards.txt to get your board running with the upstream repo or minimized modification in your folk. 

However, you really need to break this PR into multiple sub one, I can see
- nrfx update can be on its own pr
- skip irrelevant modification to the actual SDv7
- separate PR for S113, I could give it a try.

hathach:
thank you for your patient, nrf52833 is supported by https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/739 , if that misses anything, feel free to make following up PR.

PR #601, opened 2020-11-11
replace add-path with GITHUB_PATH env variable

jpconstantineau:
replacing the add-path from github actions script so that CI keeps working...

message from github actions:
The `add-path` command is deprecated and will be disabled on November 16th. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


hathach (approved):
thank you for the PR, look good

PR #604, opened 2020-12-04
Event dispatch bugfix

orrmany:
Hi, here comes a bugfix for a missing dereference. Combined with the headache of migration to PlatformIO Core 5.x, this caused me several hours frustration.. 
Probably I made the original erroneous pull-request into 0.21.0 from the beginning....

### DIFF
--- a/libraries/Bluefruit52Lib/src/bluefruit.h
+++ b/libraries/Bluefruit52Lib/src/bluefruit.h
@@ -177,9 +177,9 @@ class AdafruitBluefruit
      * Optional semaphore for additional event handlers for SD event.
      * It can be used for handling non-BLE  SD events 
      *------------------------------------------------------------------*/
-    void setMultiprotocolSemaphore(SemaphoreHandle_t* p_mprot_event_semaphore) 
+    void setMultiprotocolSemaphore(SemaphoreHandle_t mprot_event_semaphore) 
     { 
-        _mprot_event_sem= p_mprot_event_semaphore;
+        _mprot_event_sem= mprot_event_semaphore;
     } 
 #endif
 
@@ -229,7 +229,7 @@ class AdafruitBluefruit
     /* Optional semaphore for additional event handlers for SD event.
      * It can be used for handling non-BLE  SD events 
      */
-    SemaphoreHandle_t* _mprot_event_sem;
+    SemaphoreHandle_t _mprot_event_sem;
 #endif
 
     // Auto LED Blinky


orrmany:
@hathach, maybe we should rename `Bluefruit._mprot_event_sem` into `Bluefruit._mprot_event_sem_p` to prevent similar errors?

orrmany:
Well, I found the commit, which introduced the bug:  [ ff12326 ](https://github.com/adafruit/Adafruit_nRF52_Arduino/commit/ff12326ecf6bafa2584286a67e5372593f2fb1e3#diff-52fbf28607e4a9f883053a09cf899009a78f97f16054e847d2225ff93c483365). I made this somehow while I made editorial changes :(

hathach:
> @hathach, maybe we should rename `Bluefruit._mprot_event_sem` into `Bluefruit._mprot_event_sem_p` to prevent similar errors?

the semaphore is the pointer itself, maybe you could just use `SemaphoreHandle_t` to declare `_mprot_event_sem` and `setMultiprotocolSemaphore`

orrmany:
..but you wanted minimize the ANT support impact on Bluefruit?

If I would change the type to Semaphore_handle_t instead of
Semaphore_handle_t*, then we would need to handle the entire life-cycle
management of the semaphore inside Adafruit't Bluefruit lib, such as
calling xSemaphoreCreateBinary(), etc....

Ha Thach <notifications@github.com> ezt írta (időpont: 2020. dec. 7., H,
5:54):

hathach:
semaphore handle is pointer itself, there is no need to handle it within the bluefruit, just set(handle) when created, and set(NULL) before delete by application would work.

orrmany:
> 
> 
> semaphore handle is pointer itself, there is no need to handle it within the bluefruit, just set(handle) when created, and set(NULL) before delete by application would work.

Well, then you have to make that handle a public member, otherwise you can't invoke methods on it outside from the class.  I can't invoke `xSemaphoreCreateBinary( _mprot_event_sem)` on it, unless this is a public member. 

If this is a public member, then I can't have access control on it:  _other_ 3rd pty code could fiddle with "my" semaphore and I couldn't prevent that. There is a reason why I proposed only a setter method and not a getter method...

orrmany:
> 
> 
> semaphore handle is pointer itself, there is no need to handle it within the bluefruit, just set(handle) when created, and set(NULL) before delete by application would work.

Hmm, that would mean some API change to yet another library I am developing, but maybe your are right. I will come back to you soon

hathach:
> Hmm, that would mean some API change to yet another library I am developing, but maybe your are right. I will come back to you soon

If you don't want to change your lib, that is ok. We can leave it as it is now. Let me know what you decide to do

orrmany:
Hi @hathach, I have implemented your proposal: as part of the bugfix  I made `_mprot_event_sem` a regular semaphore

hathach (approved):
look good

PR #607, opened 2020-12-07
Science journal

hathach:
- added firmware sketch to use with Arduino Science Journal app https://www.arduino.cc/education/science-journal
- tested on CLUE, Feather nRF52840 Sense and CPB with on-board sensor.
- Everything work well: accel, gyro, magnetic, temperature, pressure, humid, proximity, color, light, sound/noise PDM. But some may need a bit sensor data tune-up (and/or calibrate etc ..)

Update: have been following closely to [Nano33BLESenseFirmware](https://github.com/arduino-libraries/Arduino_ScienceJournal/blob/master/examples/Nano33BLESenseFirmware/Nano33BLESenseFirmware.ino#L19) and support all of those sensors. But I just realize there is also LED, Input, Output characteristic as well [PhysicsLabFirmware.ino](https://github.com/arduino-libraries/Arduino_ScienceJournal/blob/master/examples/PhysicsLabFirmware/PhysicsLabFirmware.ino) , will update PR to also support these as well.

### DIFF
--- a/libraries/Bluefruit52Lib/examples/Peripheral/arduino_science_journal/arduino_science_journal.ino
+++ b/libraries/Bluefruit52Lib/examples/Peripheral/arduino_science_journal/arduino_science_journal.ino
@@ -26,6 +26,7 @@
 const int VERSION = 0x00000001;
 const float TEMPERATURE_CALIBRATION = -5.0;
 
+// 555a0002-val-467a-9538-01f0652c74e8"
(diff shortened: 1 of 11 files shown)


hathach:
update: the `PhysicsLabFirmware.ino` is meant for ARDUINO MKR WIFI 1010 and use other sets of UUID. It seems like they are not meant to work together. Therefore nothing for us to do, this PR is good for merging.

PR #608, opened 2020-12-11
Fixes MIDI read loop blocking other tasks in BLEMidi example

### LINKED ISSUE #392: blemidi example won't continue play notes
**Describe the bug**
the blemidi example won't continue playing notes through bluetooh midi. 

  - Your Board and BSP version
PCA1059, pca10059_bootloader-0.2.13_s140_6.1.1.hex
  - Your OS (mac/windows/linux) and its version
windows10
  - Serial debug log (enable IDE's Debug Mode Level to 1 or 2)  
cannot check because it happens without serial monitor

**To Reproduce**
1. open and upload blemidi example into pca1059 board.
2.  connect it to garageband app of ipad through bluetooth midi protocol.
3. the garageband receives no note from the blemidi example as expected.
4. but if we open a serial monitor (for example, the integrated one of arduino IDE), it will play a note. 
5. click the send button of the serial monitor will play a note once a time. 
6. even closing the monitor will play a note. But it will never continue.

**Expected behavior**
I expect it continue playing note standalone without serial monitor or even USB connected.

spacenate:
Fixes #392

As described in that issue, the Arduino Scheduler seems to require that loops `delay()` or `yield()` periodically, otherwise other tasks may be blocked. Constructive criticism is welcome.

### DIFF
--- a/libraries/Bluefruit52Lib/examples/Peripheral/blemidi/blemidi.ino
+++ b/libraries/Bluefruit52Lib/examples/Peripheral/blemidi/blemidi.ino
@@ -174,5 +174,8 @@ void midiRead()
 
   // read any new MIDI messages
   MIDI.read();
+
+  // Allow the Scheduler to run other tasks
+  yield();
 }
 


spacenate:
Upon re-testing, I discovered the issue I was experiencing is not caused by a `yield()` (or lack thereof), but rather with a specific MIDI application (Xequence for iOS).

PR #610, opened 2020-12-12
Update itoa.c for -std=c++17 compatibility

AnthonyDiGirolamo:
This change copies over itoa.h and itoa.c from https://github.com/adafruit/ArduinoCore-samd/tree/master/cores/arduino

This adds in the missing utoa and itoa functions when compiling using -std=c++17. This nRF52 core is currently using -std=gnu++11 in platform.txt so this isn't normally an issue.

Tested compiling a few example sketches successfully in the Arduino IDE with this change. Not tested on device, haven't ordered hardware yet.

I'm hoping to get this fixed to add support for building pigweed on top
of this core. See https://pigweed.dev/targets/arduino/target_docs.html if interested! :smile:

### DIFF
--- a/cores/nRF5/itoa.c
+++ b/cores/nRF5/itoa.c
@@ -23,6 +23,55 @@
 extern "C" {
 #endif
 
+/* reverse:  reverse string s in place */
+/*
+static void reverse( char s[] )
+{
+  int i, j ;
+  char c ;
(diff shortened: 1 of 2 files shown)


hathach (approved):
Thank you for the pr, it looks good.

PR #613, opened 2021-01-02
Endian-swap implemented in setMajorMinor

ErikHougaard:
The Endian-swap was missing on the BLEBeacon::setMajorMinor function. I was considering moving it into the start function, but as it was originally implemented in the constructor, I decided to add it to setMajorMinor instead.


hathach (approved):
yeah, totally forgot the swap. Thank you for your PR.

PR #619, opened 2021-01-27
Add CircuitPython BLE AdafruitColor example

alexwhittemore:
CircuitPython has a whole structure around communicating data with advertisements that's very easy to use. I personally needed to move my peripheral device firmware to Arduino, but wasn't sure exactly how to format arbitrary data into an advertisement in such a format that it would be compatible with CircuitPython `Advertisement` subclasses. 

This sketch demonstrates the Bluefruit.Advertising API(). When powered up, the Bluefruit module will start advertising a packet compatible with the AdafruitColor class from CircuitPython (https://github.com/adafruit/Adafruit_CircuitPython_BLE/blob/master/adafruit_ble/advertising/adafruit.py)
 
In other words, `ble.start_scan(AdafruitColor)` in CircuitPython will hear an advertisement from this device containing the color 0x0F0033 (violet).

Presumably, anyone clever enough to make their own `Advertisement` subclass in CircuitPython will have no trouble adapting this example on the Arduino side to match.


hathach (approved):
Thank you for your pr. Sorry for being late, I was a bit busy with other works

PR #627, opened 2021-03-30
add generic nrf52833 support

braingram:
I had a need to get this working on a custom nrf52833 board and wanted to share the result. I don't have a pca10100 to test with but could get one if it would prove useful (and add that board to the core) to validate this pull request. Otherwise I added a 'generic nrf52833'.

I mostly pulled files over from the nrfx-v2.1.0 mdk but did make some changes to the gcc_startup assembly to make it fill the stack with 0xAD like the other chips.


hathach (changes requested):
Thanks for your PR, though generic board is troublesome for future support and maintenance. Everything is too vague and people could use different gpio for same purpose with all sort of confusion. Please pick one of well known board such as pca10010 with correct pin mapping. PCA10100 has standard UNO header, which will help to test with other shield/sensor etc ...  The changes look good though, 833 is not too different from 840 except for the crypto cell cc310

LokiMetaSmith:
I applied these changes and needed to add a few more to get bluefruit support to compile. 

I also added in the changes needed to get DSP to compile. I know you're not suppose to combine multiple features like this, but I'm still figuring out git and wanted to share what I have working. 

https://github.com/LokiMetaSmith/Adafruit_nRF52_Arduino/tree/ada_master

hathach:
@LokiMetaSmith I have DSP lib already enabled locally (also update CMSIS to 5.7.0) on effort to compiled with latest tensorflow. You should just focus on the port.

PR #642, opened 2021-05-20
updated linker script to fix recurring issue with .bss linker warning

jpconstantineau:
This update to the "nrf52_common.ld" linker script fixes the linker warning **warning: changing start of section .bss by 4 bytes** which is given to sketches that contain lots of static/global data.


hathach (approved):
thank you for the PR, look great to me. Github ci seem to have some issue. Will merge when ci passed.

PR #651, opened 2021-05-27
Added the makerdiary pitaya go board

y4m-y4m:
This PR adds the makerdiary pitaya go board: https://github.com/makerdiary/pitaya-go
Sketches upload to the board, analog and digital pins work. Wifi via the WiFi101 library also works. I haven't tested NFC because I'm not sure how. There are still some problems for example when I'm using Serial in my sketch, the COM port in the Arduino IDE displays the board name as "Adafruit Feather nRF52840 Express", no idea how to change that. Also I'm not sure what I should be using for `pitayago.build.board=NRF52840_FEATHER` in boards.txt. Any and all suggestions for improvement are welcome.


hathach:
hi, pitaya need to fork and create their own board support package - Adafruit does not manufacture it so we wouldn't be able to add support. thanks :)

PR #657, opened 2021-06-05
Fix to atomic compiler errors

jpconstantineau:
This simple fix is to resolve the issue with compilation errors on "atomic_uint32_t".  This is an alias for std::atomic<std::uint32_t>.  When replacing it with the unaliased type, the errors are gone.


hathach (approved):
look good, thanks for the PR

PR #675, opened 2021-07-13
Allow common_func.h to compile without error when it is included inside extern "C"

pyro9:
When a library includes something that includes common_func.h from inside an extern "C" {} block, compilation will fail. This is at least an infelicity. This PR forces C++ linking in the C++ only parts of common_func.h allowing it to compile without mysterious failures.

### DIFF
--- a/cores/nRF5/common_func.h
+++ b/cores/nRF5/common_func.h
@@ -39,6 +39,7 @@
 
 
 #ifdef __cplusplus
+  extern "C++" {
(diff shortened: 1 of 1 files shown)


hathach (approved):
This is really bizarre situation. Though this is legit. I feel that this is over complicated for `arrcount()`. I will try to simplify these implementation later on, maybe just keep the C version.

PR #683, opened 2021-08-04
Add SAADC offset calibration and reading of CPU temperature

ajs123:
These additions

- add SAADC offset calibration to the analog library. Nordic recommends calibration if the system temperature is expected to change by more than 10 degrees C.
- add the ability to read the CPU (chip) temperature.

Notes:
- readCPUTemperature() returns a floating point value. Internally, the TEMP peripheral provides an integer with 1 LSB = 0.25 degrees C. Repeatability is +/- 0.25 C but accuracy is only +/- 5 C.  The floating point value is a convenience for the novice user. If just passing along the integer is preferred, I can change it.
- Though initiating offset calibration is probably the main reason for checking the temperature, readCPUTemperature() was added to wiring.c/h rather than to wiring_analog_. I'd be happy to move it if another location is preferred. I didn't see any more logical place, and it didn't seem worthwhile to create any new files.
- Though existing header files include #defined values for flag and trigger fields, both functions used hard-coded constants of 0x00 for not ready or disable, nonzero for ready, and 0x01 for triggers. This follows the existing pattern as well as many Nordic examples, so it is likely to be reasonably future-proof.
- I left in a few comments that perhaps should be removed.

### DIFF
--- a/cores/nRF5/wiring.c
+++ b/cores/nRF5/wiring.c
@@ -147,22 +147,6 @@ void systemOff(uint32_t pin, uint8_t wake_logic)
 
 float readCPUTemperature()
 {
-  // These are defined simply to make the ensuing code easier to read. Ready and triggers are 0x01, and not ready is 0x00. 
-  // Many other library functions use a literal 0x00 to reset flags and a literal 0x01 for triggers, and test "done"
-  // flags for nonzero.  In this function, we used the defined constants. Using 0x00 and 0x01 would probably be fine
-  // because they're unlikely to ever change.
-  const uint32_t temp_ready = ( (TEMP_EVENTS_DATARDY_EVENTS_DATARDY_Generated << TEMP_EVENTS_DATARDY_EVENTS_DATARDY_Pos)
-                                && TEMP_EVENTS_DATARDY_EVENTS_DATARDY_Msk );
-
-  const uint32_t temp_not_ready = ( (TEMP_EVENTS_DATARDY_EVENTS_DATARDY_NotGenerated << TEMP_EVENTS_DATARDY_EVENTS_DATARDY_Pos)
-                                    && TEMP_EVENTS_DATARDY_EVENTS_DATARDY_Msk );
-
-  const uint32_t temp_trigger = ( (TEMP_TASKS_START_TASKS_START_Trigger << TEMP_TASKS_STOP_TASKS_STOP_Pos)
-                                  && TEMP_TASKS_START_TASKS_START_Msk );
-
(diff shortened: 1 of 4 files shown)


cores/nRF5/wiring_analog_nRF52.c
```diff
@@ -284,6 +284,25 @@ uint32_t analogReadVDD( void )
   return analogRead_internal(SAADC_CH_PSELP_PSELP_VDD);
 }
 
+analogCalibrateOffset()
```
hathach, line 287: would you mind adding the return type and argument explicitly to void

```C
void analogCalibrateOffset(void)
```

cores/nRF5/wiring.c
```diff
@@ -143,3 +143,27 @@ void systemOff(uint32_t pin, uint8_t wake_logic)
     NRF_POWER->SYSTEMOFF = 1;
   }
 }
+
+
+float readCPUTemperature()
```
hathach, line 148: float is perfect, it is indeed the right choice for the function. Though can you add void as its explicit argument

```C
float readCPUTemperature(void)
```

hathach (commented):
Thank you very much for your PR. Everything look great and spot on, I only have a minor request to add void explicitly to function signature. If you are busy, I will add it later on, not a problem at all.

hathach:
> * readCPUTemperature() returns a floating point value. Internally, the TEMP peripheral provides an integer with 1 LSB = 0.25 degrees C. Repeatability is +/- 0.25 C but accuracy is only +/- 5 C.  The floating point value is a convenience for the novice user. If just passing along the integer is preferred, I can change it.

It is perfect choice, I am happy with float. M4 has FPU after all.

> * Though initiating offset calibration is probably the main reason for checking the temperature, readCPUTemperature() was added to wiring.c/h rather than to wiring_analog_. I'd be happy to move it if another location is preferred. I didn't see any more logical place, and it didn't seem worthwhile to create any new files.

Perfect location 

> * Though existing header files include #defined values for flag and trigger fields, both functions used hard-coded constants of 0x00 for not ready or disable, nonzero for ready, and 0x01 for triggers. This follows the existing pattern as well as many Nordic examples, so it is likely to be reasonably future-proof.

no problem at all, I use direct 0,1 for enable as well.

> * I left in a few comments that perhaps should be removed.

the more comments, the better it is

ajs123:
Oops! Sorry about that. I can do it tonight or in the morning.

hathach (approved):
superb! thank you very much for your PR. I am sure others will find this very useful

PR #685, opened 2021-08-18
Release 1.0.0

hathach:
- bump up version and update changelog for relase. Since the core is stable enough, I think it is good time to release as 1.0.0 since current number is already too high (0.24.0)
- Also take chances to update included TinyUSB lib to 1.4.4 and bootloader binaries to 0.6.1


jpconstantineau:
Has 1.0.0 been officially released? I don't see it in the GitHub releases or in Arduino IDE.  I would like to re-base my whole project  (BlueMicro_BLE keyboard firmware) and my Community nRF52 BSP as it's still pointing to before all the recent updates but would like to do so on something available to everyone and that wont be changing soon.  Thanks (avoided creating a whole issue for such a thing...)

PR #622, opened 2021-01-31
Arduino BLE Gamepad support

nekuneko:
This is a basic draft to get BLE Gamepad support in Adafruit nRF52 devices. I based the Gamepad code on Mouse code as both peripherals are so similar. I'm not sure about lines `BLECLientHidAdafruit.cpp:44` and` BLEHidGeneric.cpp:248` as I used `UUID16_CHR_BOOT_MOUSE_INPUT_REPORT` constant as 16 bit UUID for gamepad, but it works. Here some considerations:

- It's needed to add some little changes on `hid.h` file from `Adafruit_TinyUSB_ArduinoCore` repo. See PR #20 from `adafruit/Adafruit_TinyUSB_ArduinoCore` repo.
- This may will solve #228. 
- There is no support for HAT buttons, but I can try to implement it if needed. It would be necessary to modify again all these files.

PR #688, opened 2021-08-25
Ble gamepad

### LINKED ISSUE #36: Separate HID joystick/gamepad service

### LINKED ISSUE #228: Please add BLEHID gamepad support and example
I want to use a nRF52840 as a BLE HID gamepad, but I saw, that the gamepad descriptor is commented out in BLEHidAdafruit.cpp. Is there something wrong with the code?

If so, what is needed to fix it?

Also, does somebody have an example for a gamepad?

hathach:
This PR based on and supersede #622 to add BLEHidGamepad support as separated class from BLEHidAdafruit and blehid_gamepad example. which is inspired by @nekuneko exellent PR support for gamepad  for both usb and ble.

example is tested on android and windows. Unfortunately I don't know how to test it on iOS/iPad, maybe Apple require an MFi licence or something. If anyone know how to test this out, please let me know. 

Note: also take chance to rename example from hid_ to blehid_ (to distinguish from usb hid). Also tweaking existing example. 
close #622 , fix #228, fix #36


ThomasAtBBTF:
Will there also be an CircuitPython equivalent? And a sample?
I would be very interested in that.

EtienneGameSeed:
Hello,
I've tried the BLE GamePad example on an NRF52840 Feather Express and on 2 different Windows 10 portable PC with integrated Bluetooth. On both PC is behaves the same way:
- I add a bluetooth device, the PC sees the feather, I can click on the device to connect to it.
- I can pair the device, the game pad is correctly recognised and works as it should. The devices appears as "Connected".
- when I unplugg the Feather's power, the devices is still present on the PC, and is marked as "Paired", but the game controller is removed. This is the normal behaviour I suppose.
- when I power on the Feather again, the pairing is done, then on the PC it shows "Connected" for a short amount of time, and then disconnects. Then pairing starts again, it connects, and then disconnects again, this cycle occurs forever.
- To make it work again, I have to delete the device on the PC, and add it again.

I have tried on 2 different PCs (Windows 10), it behaves the same way.
I have tried with an ESP 32 and the ESP32 BLE gamepad library, and it works (no need to delete the device when power off/on), so the problem is not comming from the PCs Bluetooth stacks.

PR #689, opened 2021-08-31
UUID enumeration update

nosv-dr:


hathach (approved):
thank you for the PR. It has been quite some time since I updated these value

PR #659, opened 2021-06-09
Nrf52833 support

LokiMetaSmith:
adds nordic demo boards pca10059 and pca10100, also includes support for nrf52833

PR #698, opened 2021-10-16
nRF52833 support WIP

IanBallinger:
Runs on device hardware (not yet tested on pca10100). Flashing with nrfjprog 10.12.1 manually installed into BSP. 

Current issue: Bluefruit.begin() breaks when bond_init() calls InternalFS functions. 

Implementation borrows from work done by @LokiMetaSmith in PR #659.


jpconstantineau:
I am really interested in this.  I made sure that CircuitPython support was working fine for the 833 but have been hoping for Arduino support for the 833 to come.  Debugging on CP pointed me to update the bootloader SD.  The UF2 device family is also different with the 833  (see bootloader repo)

Couple of comments:
1 - are the bootloader files the updated ones?
2 - will this support the 7.x SD that's in the newly updated bootloader? (the 6.x SD isn't supported by the 833)

IanBallinger:
Per the [compatibility matrix](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fug_gsg_ses%2FUG%2Fgsg%2Fsoftdevices.html), the nRF52833 requires S140 7.0.1 or later. The current bootloader included in this fork was built with S140 6.1.1 for compatibility with the rest of the core since I think some function signatures change between versions (but i could be wrong). A softdevice update would probably be the last major thing needed to get board support working.

variants/pca10100/variant.h
```diff
@@ -0,0 +1,155 @@
+ /*
```
IanBallinger, line 1: file has not been double checked, it just follows the same structure as the other dev kit variants.

variants/pca10100/variant.cpp
```diff
@@ -0,0 +1,55 @@
+/*
```
IanBallinger, line 1: file has not been double checked, it just follows the same structure as the other dev kit variants.

cores/nRF5/nordic/nrfx_config.h
```diff
@@ -26,7 +26,7 @@
 #define NRFX_TIMER2_ENABLED 0
 #define NRFX_TIMER3_ENABLED 0
 
-#ifdef NRF52840_XXAA
+#if defined(NRF52840_XXAA)
```
IanBallinger, line 29: the nRF52833 does not have a QSPI peripheral

PR #703, opened 2021-11-24
clear STOPPED event when start transmission,

hathach:
clear STOPPED event when start transmission,  which could be caused by… missing ack from previous transaction. Part of effort to resolve
- https://github.com/adafruit/Adafruit_LTR390/issues/2

### DIFF
--- a/libraries/Wire/Wire_nRF52.cpp
+++ b/libraries/Wire/Wire_nRF52.cpp
@@ -51,13 +51,13 @@ void TwoWire::begin(void) {
   //Main Mode
   master = true;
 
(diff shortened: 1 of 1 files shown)


ladyada:
i didnt test but i can if you like! also please check LC709203 breakout - it also had issues!

hathach:
> i didnt test but i can if you like! also please check LC709203 breakout - it also had issues!

I have tested with actual LTR390, and it works well with both nrf and rp2040. I don't have the LC709203, but could make an PR as best as I could for you to test it out. Give me a few minutes to look at its code.

hathach:
is it the issue you mentioned, it seems to be different one, I will ask user for more info. But I guess an actual breakout is probably needed for testing.

PR #708, opened 2021-12-24
fix issue with blitFrameBuffer not working on Adafruit CLUE (HT @jimsbugs)

dstieglitz:
Fixes an issue identified here https://forums.adafruit.com/viewtopic.php?f=65&t=185669 where blitFrameBuffer hangs on Adafruit CLUE

### DIFF
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -207,8 +207,13 @@ void SPIClass::transfer(const void *tx_buf, void *rx_buf, size_t count)
 
   while (count)
   {
+#ifdef _VARIANT_CLUE52840_
+    // each transfer can only up to 64KB (16-bit) bytes
+    const size_t xfer_len = min(count, UINT16_MAX);
+#else
     // each transfer can only up to 64KB (16-bit) bytes
     const size_t xfer_len = min((uint16_t) count, UINT16_MAX);
+#endif
 
     nrfx_spim_xfer_desc_t xfer_desc =
     {
--- a/libraries/SPI/SPI.cpp
+++ b/libraries/SPI/SPI.cpp
@@ -207,13 +207,8 @@ void SPIClass::transfer(const void *tx_buf, void *rx_buf, size_t count)
 
   while (count)
   {
-#ifdef _VARIANT_CLUE52840_
     // each transfer can only up to 64KB (16-bit) bytes
     const size_t xfer_len = min(count, UINT16_MAX);
-#else
-    // each transfer can only up to 64KB (16-bit) bytes
-    const size_t xfer_len = min((uint16_t) count, UINT16_MAX);
-#endif
 
     nrfx_spim_xfer_desc_t xfer_desc =
     {


libraries/SPI/SPI.cpp
```diff
@@ -207,8 +207,13 @@ void SPIClass::transfer(const void *tx_buf, void *rx_buf, size_t count)
 
   while (count)
   {
+#ifdef _VARIANT_CLUE52840_
+    // each transfer can only up to 64KB (16-bit) bytes
+    const size_t xfer_len = min(count, UINT16_MAX);
+#else
     // each transfer can only up to 64KB (16-bit) bytes
     const size_t xfer_len = min((uint16_t) count, UINT16_MAX);
+#endif
```
hathach, line 216: thank for the PR, it is not really clear to me how the change would fix the bug. Would you provide which value the count that is used when it causes the issue with/without the PR.

jimsbugs:
With the `(uint16_t)` cast in place, any value for `count` exceeding 65535 causes `SPI::transfer` to go into an infinite loop.  Consider the case where `count = 65538`:

- on the first iteration, `(uint16_t)count` evaluates to `2`, so `xfer_len` becomes `min(2, 65535)`, that is, 2, so two bytes are transferred and `count` is decremented to 65536
- on the second iteration, `(uint16_t)count` evaluates to `0`, so `xfer_len` becomes `min(0, 65535)`, that is, 0, so _no_ bytes are transferred and `count` remains unchanged, and the loop never exits

Without the cast, we get the following:

- on the first iteration, `count` evaluates to `65538`, so `xfer_len` becomes `min(65538, 65535)`, that is, 65535, so 65535 bytes are transferred and `count` is decremented to 3
- on the second iteration, `count` evaluates to `3`, so `xfer_len` becomes `min(3, 65535)`, that is, 3, so the final three bytes are transferred and `count` is decremented to 0 and the loop exits

The reason we see this on the CLUE is its relatively large 240x240 screen - a full screen blit is 240x240x2=115200 bytes.  With the `(uint16_t)` cast, the first iteration of the loop transfers 49664 bytes, then the loop gets stuck, and the CLUE hangs.  Without the cast, the first iteration transfers 65535 bytes, the second 49665 bytes, and then the loop exits.  The bug itself, of course, is not specific to the CLUE (I probably should have mentioned that on the Adafruit forum - sorry @dstieglitz).

My apologies if this is too much detail - the code with the cast does not look obviously wrong and it took me some time to understand it and come up with a fix.  Hopefully this helps.

libraries/SPI/SPI.cpp
jimsbugs: With the `(uint16_t)` cast in place, any value for `count` exceeding 65535 causes `SPI::transfer` to go into an infinite loop.  Consider the case where `count = 65538`:

- on the first iteration, `(uint16_t)count` evaluates to `2`, so `xfer_len` becomes `min(2, 65535)`, that is, 2, so two bytes are transferred and `count` is decremented to 65536
- on the second iteration, `(uint16_t)count` evaluates to `0`, so `xfer_len` becomes `min(0, 65535)`, that is, 0, so _no_ bytes are transferred and `count` remains unchanged, and the loop never exits

Without the cast, we get the following:

- on the first iteration, `count` evaluates to `65538`, so `xfer_len` becomes `min(65538, 65535)`, that is, 65535, so 65535 bytes are transferred and `count` is decremented to 3
- on the second iteration, `count` evaluates to `3`, so `xfer_len` becomes `min(3, 65535)`, that is, 3, so the final three bytes are transferred and `count` is decremented to 0 and the loop exits

The reason we see this on the CLUE is its relatively large 240x240 screen - a full screen blit is 240x240x2=115200 bytes.  With the `(uint16_t)` cast, the first iteration of the loop transfers 49664 bytes, then the loop gets stuck, and the CLUE hangs.  Without the cast, the first iteration transfers 65535 bytes, the second 49665 bytes, and then the loop exits.  The bug itself, of course, is not specific to the CLUE (I probably should have mentioned that on the Adafruit forum - sorry @dstieglitz).

My apologies if this is too much detail - the code with the cast does not look obviously wrong and it took me some time to understand it and come up with a fix.  Hopefully this helps.

hathach (commented):
excellent and very detailed explanation. Thank you very much for the PR. Since this issue affects all variants, so we should apply the change for all, would you mind push update to drop the cast entirely for all variants.

```C
    // each transfer can only up to 64KB (16-bit) bytes
    const size_t xfer_len = min(count, UINT16_MAX);
```

hathach (approved):
thank you, look great, will merge after ci passed

dstieglitz:
Thanks for the excellent fix/explanation @jimsbugs

PR #735, opened 2022-06-24
Particle Argon variant

ericbieri:
Hi, I'd like to contribute the Particle Argon variant. Please review.


ericbieri:
Hi hatach, that's disappointing. I bought the Argon from Adafruit about two years ago and it is still currently on sale at Adafruit (https://www.adafruit.com/product/3993). I even pointed out in the "Readme.md" that the Argon is not supported (same as Xenon) by Adafruit. What is the reason for this decision? The Xenon was also integrated, why not the rest of the family (Argon & Boron)? Can you not make an exception?

ladyada:
we don't manufacture the board and so we cannot effectively manage or maintain support, folks are welcome to fork and make 3rd party board support packages. we need to be able to focus on supporting adafruit-made boards :)

ericbieri:
Thanks for your feedback and keep up the great work! :-)

Issue #737, opened 2022-07-07
Building for nRF52832 suddenly fails without changes when including bluefruit.h 


CSC-Sendance:
### Operating System

Windows 10

### IDE version

PlatformIO Core 6.1.0 Home 3.4.2

### Board

nRF52832 Bluefruit Feather

### BSP version

release

### Sketch

```
; PlatformIO Project Configuration File

[env:adafruit_feather_nrf52832]
platform = nordicnrf52
board = adafruit_feather_nrf52832
framework = arduino

```

### What happened ?

Building a platformio project aiming with adafruit_feather_nrf52832 as target board, nordicnrf52 platform, and arduino framework , suddenly fails with a message `  #error CryptoCell CC310 is not supported on this board`   on compiling, when including <bluefruit.h>. This was not the case yesterday and also isn't on colleague's PCs. 

This part from Adafruit_nRFCrypto.cpp causes this message:
`// Only nRF52840 has CC310
#ifndef NRF_CRYPTOCELL
  #error CryptoCell CC310 is not supported on this board
#endif`

NRF_CRYPTOCELL is defined in nrf52840.h (there also appears to be no nrf52832.h) .

I would guess that this may be caused by an automatic version bump of some library but haven't found any particularly recent releases or changes in, e.g. the nrfCrypto library.

### How to reproduce ?

1. Make a new platformio project with "adafruit_feather_nrf52832" as target board
2. add `#include bluefruit.h` in main.cpp 
3. Try to build

### Debug Log

```

> Executing task: C:\.platformio\penv\Scripts\platformio.exe run --environment adafruit_feather_nrf52832 <

Processing adafruit_feather_nrf52832 (platform: nordicnrf52; board: adafruit_feather_nrf52832; framework: arduino)
----------------------------------------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/nordicnrf52/adafruit_feather_nrf52832.html
PLATFORM: Nordic nRF52 (9.4.0) > Adafruit Bluefruit nRF52832 Feather
HARDWARE: NRF52832 64MHz, 64KB RAM, 512KB Flash
DEBUG: Current (blackmagic) External (blackmagic, cmsis-dap, jlink, stlink)
PACKAGES:
 - framework-arduinoadafruitnrf52 @ 1.10300.0 (1.3.0)
[...]

PR #738, opened 2022-07-08
Fix wrongful inclusion of Adafruit nRFCrypto library

maxgerhardt:
Fixes https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/737

PlatformIO's default LDF search mode sees the code at 

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/10ed826056f9c0b3079129a3ec37c3a90ecb0585/libraries/Bluefruit52Lib/src/BLESecurity.h#L31-L33

and always includes the crypto lib.

Fix is turning up the [LDF mode](https://docs.platformio.org/en/latest/projectconf/section_env_library.html#lib-ldf-mode) ([2](https://docs.platformio.org/en/latest/librarymanager/ldf.html#dependency-finder-mode)) to one that evaluates the macro correctly and prevents the inclusion of the library when `NRF_CRYPTOCELL` is not defined for boards that don't have it.

I still have to test this..

### DIFF
--- a/libraries/Bluefruit52Lib/library.json
+++ b/libraries/Bluefruit52Lib/library.json
@@ -1,0 +1,25 @@
+{
+  "name": "Adafruit Bluefruit nRF52 Libraries",
+  "version": "0.21.0",
+  "description": "Arduino library for nRF52-based Adafruit Bluefruit LE modules",
+  "keywords": "bluefruit, ble",
+  "repository":
+  {
+    "type": "git",
+    "url": "https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/master/libraries/Bluefruit52Lib"
+  },
+  "authors":
+  [
+    {
+      "name": "Adafruit",
+      "email": "info@adafruit.com",
+    }
+  ],
+  "license": "BSD-3",
+  "frameworks": "arduino",
+  "platforms": "nordicnrf52",
+  "build":
+  {
+    "libLDFMode": "chain+"
+  }
+}


ladyada:
we really really do not want to do this, then we have to maintain two library files.

@ivankravets is there a way to fix this so platformio stops misparsing C files, maybe a different preparser? this issue keeps coming up where valid C code that has ifdefs doesnt work with platformio. or have some inline alert to tell platformio. thank you!

ladyada (changes requested):
we do not want two places that need version updates

maxgerhardt:
I understand that two version fields is bad -- but I still wanted to try to just have
```json
{
  "build":
  {
    "libLDFMode": "chain+"
  }
}
```
so that.. maybe.. PlatformIO gets version info from `library.properties` and this critical build info from the `library.json`, hence no double version maintaining.

ladyada:
yes that would be ideal... or have some extra line in the library.properties file

ivankravets:
PlatformIO Core already enables "Arduino Preprocessing" for a library if it has a `depends` field. The problem is linked with the version of this Arduino Core that is hosted in our registry. It has patched libraries. See the contents of `~/.platformio/packages/framework-arduinoadafruitnrf52/libraries/Bluefruit52Lib`.

@maxgerhardt, please move this issue to https://github.com/platformio/platform-nordicnrf52/issues

P.S: See `~/.platformio/packages/framework-arduinoadafruitnrf52/libraries/Bluefruit52Lib/extra_script.py`:

```python

# The only purpose of this extra script is to set the `NRF_CRYPTOCELL` macro for the
# nrf52840 series because the original `NRF_CRYPTOCELL` macro is set in the framework
# sources and not visible when PlatformIO LDF resolves dependencies.
if env.BoardConfig().get("build.mcu", "").startswith("nrf5284"):
    env.Append(
        CPPDEFINES=[("NRF_CRYPTOCELL", "((NRF_CRYPTOCELL_Type*) NRF_CRYPTOCELL_BASE)")],

    )
```

PR #739, opened 2022-07-14
Add support for nrf52833

hathach:
- add feather 52833 support in advance (comment out), add pca10100 as 833 for test board
- add s140 v7.3.0 API header
- update makeboards.py (thus boards.txt) and platform.txt to support 833 variant
- update bootloader binaries to 0.9.1 (to adopt 833)
- add linker, startup for 833
- update repo wide to support 833 (ifdef NRF52833_XXAA)

NOTE: nRF52833 does not have QSPI peripherals, therefore current external flash is not supported/ported yet to SPI. Will do later when the board is in production


Noteolvides:
Any news about the integration of the 52833?

PR #744, opened 2022-10-04
Implemented thread-safe realloc and calloc

godario:
Hello everybody.
I have noticed that your FreeRTOS implementation only provides the threadsafe version of `malloc` and `free` while `realloc` and `calloc` are still not threadsafe. In case of very intense use of objects like `String` (which uses `realloc`) or in general of `realloc`/`calloc`, I noticed some hangs that I solved by implementing `pvPortRealloc` and `pvPortCalloc` and then wrapping the real `realloc` and `calloc` from the linker.
Hope it can help!


hathach (approved):
look great, thank you, and sorry for the late review

PR #747, opened 2022-10-23
nrfx_temp implemented

neotje:
Hi,
I have added the source code for the nrfx_temp driver and enabled it in the nrfx_config.


hathach (approved):
look great, thank you for your PR. I make some changes to
- rename example and move it to BluefruitLib/Examples/Hardware which is for hw peripheral examples
- change the value to float for readibility

PR #759, opened 2023-02-13
Add WisBlock RAK4631 nRF52 + SX1262 LoRa module

beegee-tokyo:
I hope you can accept another third party board in your great BSP.


hathach:
unforutnately, we don't have the hardware and couldn't provide support for other vendor.

PR #766, opened 2023-04-02
Allow/Reject Pairing requests

bigfoot38:
Hi,

I'm working on a board that does not have any Yes/No Button nor Display to accept or reject bonding requests.

I wanted to allow pairing based on a time window, which opens either on reset or when a dedicated button is pressed. This behaviour is common on several bluetooth hardwares. Well I could not figure out how to do that with the actual `BLESecurity` implementation.

Consequently, I added the `BLESecurity::setPairable` method. It controls an underlying flag. When set (default value) all requests are accepted, when `false`, they are all rejected.

I hope it helps, otherwise I'm opened to discussion.

Best regards,
Bigfoot38

### DIFF
--- a/libraries/Bluefruit52Lib/src/BLESecurity.cpp
+++ b/libraries/Bluefruit52Lib/src/BLESecurity.cpp
@@ -90,6 +90,7 @@ BLESecurity::BLESecurity(void)
   _passkey_req_cb = NULL;
   _complete_cb = NULL;
   _secured_cb = NULL;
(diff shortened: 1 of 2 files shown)


hathach (changes requested):
PR does not make sense to me. you can implement your own timestamp and allow/reject it based on the timestamp. Pairing is rathe complicated, checkout pairing examples and/or Bluetooth LE docs to find the best method for your device.

PR #769, opened 2023-04-24
Add support for Immediate Alert Service

qubitter:
This PR modifies `bluefruit.h` and adds files to support instantiation and detection of the Immediate Alert service, or IAS. This service only has one characteristic - the 'alert level'.


hathach (approved):
look great, thank you very much

PR #786, opened 2023-08-24
Attempt to add PHY coded support to Adafruit API

n-snyder:
This change attempts to update the Bluefruit52 library to enable access to a few of Nordic's API settings with the hope of using BLE long-range advertising (PHY coded).  I applied these changes to an Adafruit ItsyBitsy nRF52840 Express board, attempted to advertise, and was unsuccessful with having a BT sniffer pick up the advertisements. Therefore, please review the changes well and let me know if there is an error.

I am using a second Adafruit ItsyBitsy nRF52840 Express with Wireshark to serve as the sniffer. The nRF Sniffer plug-in is set to pick up PHY coded advertisements. I share this because I cannot isolate the root cause with the hardware that I have available:
1. The board may not be capable of PHY coded advertisements and/or sniffing
2. nRF's software may not be correctly sniffing PHY coded
3. The code in this pull request is wrong
4. My application's use of the API is wrong.

If someone has enabled this functionality on one of the boards, please consider sharing the settings used.

### DIFF
--- a/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
+++ b/libraries/Bluefruit52Lib/src/BLEAdvertising.cpp
@@ -43,11 +43,12 @@ BLEAdvertisingData::BLEAdvertisingData(void)
 {
   _count = 0;
   arrclr(_data);
+  _max_data_size = BLE_GAP_ADV_SET_DATA_SIZE_MAX;
 }
 
 bool BLEAdvertisingData::addData(uint8_t type, const void* data, uint8_t len)
 {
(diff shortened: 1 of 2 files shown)


mharper (commented):
@n-snyder I don't think that abbreviating `events` to `evts` really buys you anything here. Might as well just spell it out for clarity.

hathach:
currently, I don't have the bandwidth to test and update this PR. Since it is not working yet, I convert it to draft, feel free to mark ready when it works as expected.

n-snyder:
@hathach, with further testing, this is not working yet. It is a step in the right direction, but the primary physical layer is still advertising at 1 M, rather than coded.

PR #788, opened 2023-09-05
Added support for directed BLE advertisements.

UUSim:
Can be used for example with advertising type BLE_GAP_ADV_TYPE_CONNECTABLE_NONSCANNABLE_DIRECTED.


hathach (approved):
thank you for your PR, I have tweaked the code to auto auto use peerAddress if adv type is set to one of directed type. Which is more consistent, will merge when ci passed

PR #791, opened 2023-09-21
Fix _irq_handler

rmccaskill:
I'm working on a project that has 3 rotary encoders. When I used this libary, I couldn't get the 3 switches to properly increment / decrement the steps when I called readAbs(). Looking back at the code, I see there was a change in 351cd82 which seemed to have caused the issue.

If I look previous to 351cd82, the _irq_handler was using digitalRead. I think switching to bitRead caused a mix up between the physical pin number vs. the port pin? I'm not entirely sure. I reimplemented the pre 351cd82 code and used the new ada_callback. Seems to be working great now!


hathach (approved):
changes look great, thank you for the fix. Strangely ci failed, but it isn't caused by this PR, probably due to other libraries update. Aprroved, but give me a bit of time to check out what is wrong, before merging.

PR #809, opened 2024-01-30
Line 199, BLEScanner::checkReportForUuid -- Remove const specifier fr…

ScottCase-Stemco:
…om uuid variable.

The local variable uuid in function bool BLEScanner::checkReportForUuid was defined as const, but it was not initialized with a value and is later assigned one (on lines 210 or 216). A const variable must be declared with an initial value, which cannot be changed. What is needed here is not a const value, but a cast where uuid was assigned a value. This was done correctly in one spot, but not on the other.


hathach (commented):
You probably misunderstand the purpose, it is pointer to "const memory", the pointer itself is not const.

ScottCase-Stemco:
Thanks for the explanation, hathatch!

PR #814, opened 2024-03-15
Add support for nrf52833 cores to the BSP

jeffc:
This adds support for the nrf52833 to the BSP, including the linker settings, softdevice 7.2.0 and 7.3.0 headers (to match what's available in the [bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader/tree/master/lib/softdevice)), and updates to the included libraries where appropriate.

This does **not** add any board definitions, but does include scaffolding in the `makeboards.py` to allow for new nrf52833 boards to be added based on these changes.

This is loosely based on #659


ndoo:
@ladyada @hathach sorry to bump this PR but I'm hoping this can be reviewed and pushed as it's relatively straightforward of a review (no regressions by the looks of it).

This would add native support for [Seeed XIAO BLE](https://wiki.seeedstudio.com/XIAO_BLE/) which is an s140 7.3.0 device in Arduino IDE and downstream projects like PlatformIO.

Personally, I noticed this issue because Meshtastic firmware builds using PlatformIO which uses this repo as its upstream for nRF52 support, and right now we have to produce builds by hand by manually adding the files from 691e478 of this PR - https://github.com/meshtastic/firmware/tree/master/variants/xiao_ble

ladyada:
sure, @hathach can take a look when they have time next

ndoo:
Hi @hathach, could I please request you to take a look at adding the updated softdevice versions?

hathach:
nrf52833 is just supported by https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/739, tested with pca10100, we have it on draft but didn't have time to finish it previously. To make it easier to maintain, we will only support SD v7.3, there is probably not much of difference between v7.2.0 and v7.3.0. Thank you for the PR. #739 misses anything, feel free to submit an PR

PR #822, opened 2024-05-09
Feather840 rev e

hathach:
- add PIN_NEOPIXEL_POWER as D34 for Feather nRF52840 Express RevE. Also set it as output, and write to HIGH as default
- update bootloader to 0.9.0
- bump up version to 1.6.1

@ladyada All the pin from D0-> D33 are already in use, the NEOPIXEL is added as the D34


ladyada:
thanks this works great can you do a release?

PR #826, opened 2024-06-14
Add support for VDDHDIV5

mattallen37:
Allow reading the VDDHDIV5 ADC source on devices that support it, such as nRF52840. For designs where the VDDH pin is connected to a battery, this is a great way to measure battery voltage without any additional external components.


hathach (approved):
look good, thank you

PR #827, opened 2024-07-12
Fix memory leak on failure to open

attermann:
Fixes leak of memory caused by failure to free file and directory structures when open fails.


hathach (approved):
look good, thank you

PS: ci failed with nrf52832 due to issue with tinyusb library, fixed in other repo. Should be safe to merge

PR #828, opened 2024-07-12
Making new/delete operators overridable

attermann:
Added `__attribute__((weak))` weak linking annotation to new and delete operator overrides to allow for user-defined overrides.


hathach (approved):
look good, thank you. CI failed with 832 due to tinyusb issue

PR #829, opened 2024-07-15
Null termination of file name when max length

attermann:
File names are returned unterminated in the edge case where their length is max size This is due to the fact that strncpy does terminate the destination string when the source string is greater than or equal to the specifid maximum length.


hathach (approved):
thank you

PR #831, opened 2024-08-14
Update to LFS v1.7.2 tag - to fix filesystem corruption due to file renames

geeksville:
(from https://github.com/meshtastic/firmware/issues/4395 )

Hi @hathach long time no see - I hope all is well for you!  I'm back working on Meshtastic again and one of the things in my queue was addressing some problems we've seen in the field with LFS on nrf52.  This PR is to address one of those problems - I'm submitting it upstream in hopes that it is useful for other adafruit users and also so we can stop using a temporary fork which contains this change.

## Filesystem corruption due to LFS renames

While stress testing (rarely - a couple of times a day on a node repeatedly writing/renaming/rebooting/reading files) filesystem corruption was seen that was not caused due to a reboot while writing.
After some debugging and checking github history, it seems that someone else had already found and fixed this bug:
https://github.com/littlefs-project/littlefs/commit/97d8d5e96a7781596708664f18f2ea6c3a179330

As poor luck would have it they fixed it a few **days** after adafruit forked from the master LFS tree.  The version in adafruit was forked on oct 15,2018 https://github.com/adafruit/Adafruit_nRF52_Arduino/commit/f32868509b8075366428305c860edf576499b6f1.  This version is version in framework-arduinoadafruitnrf52 is: commit cb62bf2188854c5b7c44383571ebb19a414e6137 (tag: v1.6.1)

With this fix the 'rename related' corruption no longer occurs.  

Since it seemed I needed to update LFS, I went ahead and updated it to the 'final' LFS 1.x release (v1.7.2) so that hopefully this update will be more useful to other adafruit users.  

The only other change in in v1.7.2 vs v1.6.1 is a check for 2GB files. which is not a real problem for adafruit obviously - but I included it anyways because tiny and then it will be easy for downstream users to know 'oh the version in adafruit is the final version of LFS 1.x.

I updated lfs.c and lfs.h by first generating a patch file between the old adafruit tree and v1.6.1 of LFS, then copied in the v1.7.2 version of those three files, then reapplied the patch file so that the (small number of) existing adafruit local changes were preserved.

```
adafruit specific changes to LFS are capured now in this patch file in ~/tmp
diff -crB ~/development/meshtastic/littlefs littlefs > dfile.patch
kevinh@kdesktop:~/.platformio/packages/framework-arduinoadafruitnrf52/libraries/Adafruit_LittleFS/src$ code dfile.patch 
kevinh@kdesktop:~/.platformio/packages/framework-arduinoadafruitnrf52/libraries/Adafruit_LittleFS/src$ mv dfile.patch ~/tmp/lfs-adafruit.patch
apply patch with "patch -p0 --dry-run < dfile.patch"
```

### DIFF
--- a/libraries/Adafruit_LittleFS/src/littlefs/lfs.c
+++ b/libraries/Adafruit_LittleFS/src/littlefs/lfs.c
@@ -888,7 +888,7 @@ nextname:
         }
 
         // check that entry has not been moved
-        if (entry->d.type & 0x80) {
+        if (!lfs->moving && entry->d.type & 0x80) {
             int moved = lfs_moved(lfs, &entry->d.u);
             if (moved < 0 || moved) {
                 return (moved < 0) ? moved : LFS_ERR_NOENT;
@@ -1644,6 +1644,11 @@ lfs_ssize_t lfs_file_write(lfs_t *lfs, lfs_file_t *file,
         file->pos = file->size;
     }
 
+    if (file->pos + size > LFS_FILE_MAX) {
+        // larger than file limit?
+        return LFS_ERR_FBIG;
+    }
+
     if (!(file->flags & LFS_F_WRITING) && file->pos > file->size) {
         // fill with zeros
         lfs_off_t pos = file->pos;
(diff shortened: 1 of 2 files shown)


geeksville:
btw https://github.com/meshtastic/firmware/issues/4447 might be interesting for you as well.  I'm sure you had no other option (given the tiny number of free 4KB flash pages available in the nrf52) and you probably understood the implied risk.   But I thought I'd mention it just in case it is useful to you.  Keep rocking it!

hathach (approved):
long time no see, I am sorry for super late response, I have been busy with too much of works. Thank you very much for this great findout.

geeksville:
no problem.  thanks for your work!

PR #838, opened 2025-01-17
Avoid InternalFileSystem corruption caused by simultaneous BLE operation

todd-herbert:
Unexpected BLE disconnections (such as 0x8 `BLE_HCI_CONNECTION_TIMEOUT)` cause `sd_flash_page_erase` and `sd_flash_write` operations to fail. This failure is reported with `NRF_EVT_FLASH_OPERATION_ERROR`. Currently, the InternalFileSystem library doesn't detect these events.

This PR aims to detect `NRF_EVT_FLASH_OPERATION_ERROR`, allowing several reattempts of a failed write / erase operation.

I'm unsure whether this fix is overly crude, and am concerned that I may be missing some finer detail of the filesystem implementation.  Of note is the change from a counting semaphore to a binary semaphore. Any input here would be greatly valued.

### DIFF
--- a/libraries/InternalFileSytem/src/flash/flash_nrf5x.c
+++ b/libraries/InternalFileSytem/src/flash/flash_nrf5x.c
@@ -44,11 +44,46 @@ extern uint32_t __flash_arduino_start[];
 // MACRO TYPEDEF CONSTANT ENUM DECLARATION
 //--------------------------------------------------------------------+
 static SemaphoreHandle_t _sem = NULL;
+static bool _flash_op_failed = false;
 
 void flash_nrf5x_event_cb (uint32_t event)
 {
-//  if (event != NRF_EVT_FLASH_OPERATION_SUCCESS) LOG_LV1("IFLASH", "Flash op Failed");
-  if ( _sem ) xSemaphoreGive(_sem);
+  if ( _sem ) {
+    // Record the result, for consumption by fal_erase or fal_program
+    // Used to reattempt failed operations
+    _flash_op_failed = (event == NRF_EVT_FLASH_OPERATION_ERROR);
+
+    // Signal to fal_erase or fal_program that our async flash op is now complete
+    xSemaphoreGive(_sem);
+  } 
+}
+
+// How many retry attempts when performing flash operations 
+#define MAX_RETRY 20
+
+// Check whether a flash operation was successful, or should be repeated
+static bool retry_flash_op (uint32_t op_result, bool sd_enabled) {
+  // If busy
+  if (op_result == NRF_ERROR_BUSY) {
+    delay(1);
+    return true; // Retry
+  }
+
+  // If unspecified error
+  if (op_result != NRF_SUCCESS)
+    return true; // Retry
+
+  // If the soft device is enabled, flash operations run async
+  // The callback (flash_nrf5x_event_cb) will give semaphore when the flash operation is complete
+  // The callback also checks for NRF_EVT_FLASH_OPERATION_ERROR, which is not available to us otherwise
+  if (sd_enabled) {
+    xSemaphoreTake(_sem, portMAX_DELAY);
+    if (_flash_op_failed) 
+      return true; // Retry
+  }
+
+  // Success
+  return false;
 }
 
 // Flash Abstraction Layer
@@ -107,30 +142,28 @@ static bool fal_erase (uint32_t addr)
   // Init semaphore for first call
   if ( _sem == NULL )
   {
-    _sem = xSemaphoreCreateCounting(10, 0);
+    _sem = xSemaphoreCreateBinary();
     VERIFY(_sem);
   }
 
-  // retry if busy
-  uint32_t err;
-  while ( NRF_ERROR_BUSY == (err = sd_flash_page_erase(addr / FLASH_NRF52_PAGE_SIZE)) )
-  {
-    delay(1);
-  }
-  VERIFY_STATUS(err, false);
-
-  // wait for async event if SD is enabled
+  // Check if soft device is enabled
+  // If yes, flash operations are async, so we need to wait for the callback to give the semaphore
   uint8_t sd_en = 0;
   (void) sd_softdevice_is_enabled(&sd_en);
 
-  if ( sd_en ) xSemaphoreTake(_sem, portMAX_DELAY);
-
-  return true;
+  // Make multiple attempts to erase
+  uint8_t attempt = 0;
+  while (retry_flash_op(sd_flash_page_erase(addr / FLASH_NRF52_PAGE_SIZE), sd_en)) {
+    if (++attempt > MAX_RETRY)
+      return false; // Failure
+  }
+    return true; // Success
 }
 
 static uint32_t fal_program (uint32_t dst, void const * src, uint32_t len)
 {
-  // wait for async event if SD is enabled
+  // Check if soft device is enabled
+  // If yes, flash operations are async, so we need to wait for the callback to give the semaphore
   uint8_t sd_en = 0;
   (void) sd_softdevice_is_enabled(&sd_en);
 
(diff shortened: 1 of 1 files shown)


henrygab (changes requested):
Wow!  This was a tricky bug that you tracked down.  To be clear, I don't have final say in this ... my advice here is free, and maybe you get what you pay for.

I think your solution is based on solid read of the SDK.  My comments primarily revolve around keeping the code easy to read / function naming / etc.  I hope you find it useful.

todd-herbert:
@henrygab Thanks for the feedback! I'll get onto those changes tomorrow.

todd-herbert:
[630668a](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/838/commits/630668a071e7a243689f93254c084a9162414f4c#diff-/3ea78558e12c234adba6a19bae36c214ad9d73f6L47) aims to respond to feedback from @henrygab. Hopefully I'm correctly interpreting what you're envisioning.
*  the event from the callback is now stored, rather than success / fail as a bool
* `wait_for_async_flash_op_completion` method gathers result only. Delays / queuing now handled in `fal_prog` and `fal_erase`.

Please do let me know if you feel there's any room for further improvement here, or if some new logic flaw has appeared during the refactoring.

libraries/InternalFileSytem/src/flash/flash_nrf5x.c
```diff
@@ -44,11 +45,52 @@ extern uint32_t __flash_arduino_start[];
...
+  int32_t result = initial_result;
+
+  // Operation was queued successfully
+  if (initial_result == NRF_SUCCESS) {
+
+    // Wait for result via callback 
+    xSemaphoreTake(_sem, portMAX_DELAY);
+    
+    // If completed successfully
+    if (_flash_op_result == NRF_EVT_FLASH_OPERATION_SUCCESS)
```
henrygab, line 79: Always use the curly braces.  This is particularly important in codebases that heavily use macros.  Small but important change.

libraries/InternalFileSytem/src/flash/flash_nrf5x.c
```diff
@@ -44,11 +45,52 @@ extern uint32_t __flash_arduino_start[];
...
+    xSemaphoreTake(_sem, portMAX_DELAY);
+    
+    // If completed successfully
+    if (_flash_op_result == NRF_EVT_FLASH_OPERATION_SUCCESS) 
+      result = NRF_SUCCESS;
+
+    // If general failure.
+    // The comment on NRF_EVT_FLASH_OPERATION_ERROR describes it as a timeout,
+    // so we're using a similar error when translating from NRF_SOC_EVTS type to the global NRF52 error defines
+    else if (_flash_op_result == NRF_EVT_FLASH_OPERATION_ERROR)
```
henrygab, line 85: curly braces strongly recommended.

libraries/InternalFileSytem/src/flash/flash_nrf5x.c
```diff
@@ -44,11 +45,52 @@ extern uint32_t __flash_arduino_start[];
...
+      result = NRF_SUCCESS;
+
+    // If general failure.
+    // The comment on NRF_EVT_FLASH_OPERATION_ERROR describes it as a timeout,
+    // so we're using a similar error when translating from NRF_SOC_EVTS type to the global NRF52 error defines
+    else if (_flash_op_result == NRF_EVT_FLASH_OPERATION_ERROR) 
+      result = NRF_ERROR_TIMEOUT;
+    
+    // If this assert triggers, we need to implement a new NRF_SOC_EVTS value
+    else
```
henrygab, line 89: curly braces strongly recommended (even if kept on a single line)

libraries/InternalFileSytem/src/flash/flash_nrf5x.c
```diff
@@ -140,27 +185,39 @@ static uint32_t fal_program (uint32_t dst, void const * src, uint32_t len)
...
 #if NRF52832_XXAA
-  while ( NRF_ERROR_BUSY == (err = sd_flash_write((uint32_t*) dst, (uint32_t const *) src, len/4)) )
-  {
-    delay(1);
+  // Write the page
+  // Multiple attempts, if needed
+  for (uint8_t attempt = 0; attempt < MAX_RETRY; ++attempt) {
+    err = sd_flash_write((uint32_t*) dst, (uint32_t const *) src, len/4);
+
+    if(sd_en) err = wait_for_async_flash_op_completion(err); // Only async if soft device enabled
```
henrygab, line 193: the logic appears sound.  I do recommend always using curly braces around if..else, do..while, and while blocks.   (I earned this knowledge via pain.)   Even if you keep things on one line, use the curly braces.  It avoids weird things happening when someone later uses a macro, and that macro wasn't written to be single-statement safe.

Applies also twice more below.

henrygab (commented):
Seems cleaner.  Logic remains apparently sound.  Nicely done. 🎉 

One code style comment repeated a few times.

esev:
One thing I've been wondering about this PR: If it fails after retrying, what should the behavior be. From https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/325#issuecomment-2600314556, I've seen this can lead to asserts in LittleFS. These don't necessarily happen right away and can happen some time later.

I've been wondering if the code here should assert if the retries fail. That calls attention to a possible FS corruption issue sooner, at the point where we've first detected a failure.

I've been debating this a bit myself, as it certainly isn't nice to crash if it can be avoided. But maybe it is useful in this case, to highlight the issue sooner?

Anyway, just adding this comment here in case there is a strong opinion one way or another.

henrygab:
> ... I've been wondering if the code here should assert if the retries fail. That calls attention to a possible FS corruption issue sooner, at the point where we've first detected a failure.

Those are good questions to ask.  If the corruption was guaranteed at this point, then you are right ... earlier is better, and here would prevent later-discovered inconsistencies, maybe even leave the file system in a valid state ... if never written to again.


What I'm not 100% sure of is whether a failed write is guaranteed to cause LFS corruption.  If I understand correctly, LFS is generally designed to not trust that data was actually written, just because the write reported success.  I have not dived into LFS internals for a while...

@todd-herbert ... questions for you, as you're the one who most recently dived deep....

1. Is the corruption essentially an edge case caused by the configuration choice (vs. the physical flash properties)?
2. Are there any situations where, with a similar configuration vs. physical flash, a write that fails would **_NOT_** cause LFS corruption?

Maybe, in addition to retries, since the flash is internal, changing the LFS configuration would be a worthwhile second PR for Adafruit's folks to consider?  (of course, only if it would make LFS robust to failed writes)

henrygab (dismissed):

todd-herbert:
> @todd-herbert ... questions for you, as you're the one who most recently dived deep....

I have to be honest, @esev has looked into this in much more depth than me. I've only really narrowed in on this one particular BLE disconnection case.

I'm not actually sure which situations could trigger the loop to hit `MAX_RETRY`, but maybe that's the argument in favor of asserting in this situation: to uncover any elusive edge cases which could be better handled.

> Is the corruption essentially an edge case caused by the configuration choice (vs. the physical flash properties)?

I'm no expert in the area, but reading @geeksville's thoughts in https://github.com/meshtastic/firmware/issues/4447, it does sound that the "32 LittleFS blocks per page" situation creates opportunities for corruption to occur.

henrygab:
> I'm no expert in the area, but reading @geeksville's thoughts in [meshtastic/firmware#4447](https://github.com/meshtastic/firmware/issues/4447), it does sound that the "32 LittleFS blocks per page" situation creates opportunities for corruption to occur.

I accept that reporting a failure, in at least some cases, will cause corruption in LFS.

If the assertion compiles to nothing in release builds, then sure ... this is a fine place to assert.

<details><summary>If intending release builds to lock up / crash...</summary><p/>

Concern:
* Cache layer does:  Erase(physical == 32x logical sectors @ 128 bytes each)
* Cache layer then re-writes the entire physical page
* One of those writes fails

I would not lock up on retail.  The erase has already been attempted, so nothing is saved as a result.



(+) The user experience is that the device simply hangs ... some device running tucked away in a hard-to-reach place now needs someone to go an pull the battery & power, or manually press the reset button (if exposed).  Maybe ok, maybe not.
</details> 

Because it's become clear that LFS was never designed to work as currently configured, I have little hope that any additional changes will improve the situation in a meaningful way.  Bump the retry count to **`100`**, double the delay every 20 attempts, and then crash if the operation fails.

<details><summary>A correct fix...</summary><p/>

A **_correct_** fix would be to update the LFS configuration to indicate the 4096-byte block size.  LFS supports inline'd files, so small files can end up stored inline within a sector (not taking 4k per file).  This would require testing, but something along the lines of changing:

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/4dcfa3b7960a6dafec6736a5dbea255b0946be83/libraries/InternalFileSytem/src/InternalFileSystem.cpp#L34-L35

```
#define LFS_FLASH_TOTAL_SIZE  (7*FLASH_NRF52_PAGE_SIZE) 
#define LFS_BLOCK_SIZE        FLASH_NRF52_PAGE_SIZE // more correct
```

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/4dcfa3b7960a6dafec6736a5dbea255b0946be83/libraries/InternalFileSytem/src/InternalFileSystem.cpp#L100-L119

```
   .read_size = 128; // ??? flash might support reading single byte at a time... 
   .prog_size = 128; // ??? smallest size of a write to the page ... see flash's datasheet, might be 512?
   .block_size = FLASH_NRF52_PAGE_SIZE; //  This is the physical page size
   .block_count = LFS_FLASH_TOTAL_SIZE / FLASH_NRF52_PAGE_SIZE; // e.g., seven (7)
```

The above changes are an **_ESTIMATE / STRAWMAN_** and entirely **_UNTESTED_**, as they are intended for discussion only.

</details>

todd-herbert:
> I would not lock up on retail. The erase has already been attempted, so nothing is saved as a result.

That makes sense to me.

___ 

Personally, I'd be inclined to leave this PRs scope targeting this one specific BLE disconnection issue. It seems like the fix is fairly non-controversial, and could be rolled out without too much fear of causing disruption. 

The additional discussion going on here with further aims to improve the stability of InternalFileSystem is certainly very positive and not something I'd want to discourage though!

hathach (commented):
superb ! Thank you very much for investigating and fixing the hard-to-reproduce issue. I am sure this will fix several confusing bug when radio & flash are both higly active. Thank you @henrygab for very thoughful review as usual. 

I made an attempt to tidy up the code a bit, let me knoww if it looks OK and work for you all.

todd-herbert:
Thanks for tidying it up! I've directly applied the changes from the linked text file, and it's still correctly handling BLE disconnection during flash write.

<details>
<summary>Log Output</summary>

```
DEBUG | 08:18:48 87 [Button] Opening /prefs/config.proto, fullAtomic=1
[SOC   ] NRF_EVT_FLASH_OPERATION_ERROR#
[SOC   ] NRF_EVT_FLASH_OPERATION_ERROR#
[SOC   ] NRF_EVT_FLASH_OPERATION_ERROR#
[SOC   ] NRF_EVT_FLASH_OPERATION_ERROR#
[BLE   ] BLE_GAP_EVT_DISCONNECTED : Conn Handle = 0#
[GAP   ] Disconnect Reason: CONNECTION_TIMEOUT #
INFO  | 08:18:51 91 [Button] BLE Disconnected, reason = 0x8
DEBUG | 08:18:51 91 [Button] PhoneAPI::close()
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS#
[SOC   ] NRF_EVT_FLASH_OPERATION_SUCCESS#
```

</details>

___

> PS: seems like I couldn't push to fokred PR branch, @todd-herbert could you either enable the perssion for maintainer or just apply the code here

Ah you're not the first person I've heard this from actually, although I'm not sure why it seems to happen sometimes. The "Allow edits by maintainers" box is certainly ticked.

hathach (approved):
perfect, thanks for re-testing the changes. I think your branch is forked from a forked, which complicate gh PR. I think I could use the web editor but couldn't push directly to your fork. Anyway, it is all good now.

PR #847, opened 2025-04-29
Fix a copy-paste error in the bug report template.

obra:
As I was about to file a bug, I noticed a typo, so figured it was worth sending a quick PR.


tannewt (approved):
Thank you!

PR #849, opened 2025-05-15
Adds feature reports support on Bluefruit52Lib's BLEHidGeneric service

RobertDaleSmith:
BLE HID Input and output reports are working fine but support for feature reports is missing from this lib.

WebHID Explorer offers a great tool for testing feature reports. In order to get feature reports working with the Seeed XIAO nrf Sense board, I needed to patch these changes into the platform.io framework used.

### DIFF
--- a/libraries/Bluefruit52Lib/src/services/BLEHidGeneric.cpp
+++ b/libraries/Bluefruit52Lib/src/services/BLEHidGeneric.cpp
@@ -127,6 +127,15 @@ void BLEHidGeneric::setOutputReportCallback(uint8_t reportID, BLECharacteristic:
   if ( idx < _num_output ) _chr_outputs[idx].setWriteCallback(fp);
 }
 
+void BLEHidGeneric::setFeatureReportCallback(uint8_t reportID, BLECharacteristic::write_cb_t fp)
+{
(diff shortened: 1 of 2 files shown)


libraries/Bluefruit52Lib/src/services/BLEHidGeneric.cpp
```diff
@@ -193,6 +202,21 @@ err_t BLEHidGeneric::begin(void)
     _chr_outputs[i].write8(0);
   }
 
+  // Feature reports
+  for (uint8_t i = 0; i < _num_feature; i++)
+  {
+    _chr_features[i].setUuid(UUID16_CHR_REPORT);
+    _chr_features[i].setProperties(CHR_PROPS_READ | CHR_PROPS_WRITE | CHR_PROPS_WRITE_WO_RESP);
+    _chr_features[i].setPermission(SECMODE_OPEN, SECMODE_OPEN);
```
hathach: permission is too open, should we use SECMODE_ENC_NO_MITM, SECMODE_ENC_NO_MITM as the same as output report ?

hathach (approved):
look good, thank you. I update the permission to match other hid chars.

PR #872, opened 2026-07-03
fix: TWIM EasyDMA ordering and add writeThenRead() for atomic I2C transactions

usernameLoophole:
### Summary

- Fix EasyDMA register ordering in `requestFrom()` and `endTransmission()`: DMA pointers (`RXD.PTR`/`MAXCNT`, `TXD.PTR`/`MAXCNT`) are now configured *before* triggering `TASKS_RESUME`, and a `__DMB()` barrier prevents the ARM write buffer from reordering the stores past the task trigger.
                                                                                   
- Add `writeThenRead()` — combined write-then-read that uses TWIM hardware SHORTS (`LASTTX_STARTRX` + `LASTRX_STOP`) to perform the full `START → W → REPEATED-START → R → STOP` sequence in EasyDMA without CPU involvement at the repeated-start boundary. This eliminates the START-STOP-START problem that occurs when SoftDevice or CPU preemption interrupts the software-mediated transition between `endTransmission(false)` and `requestFrom()`.

### Context
**Board: Seeed studio xiao sense ble nrf52840**
In our application we would see corrupted readings form a peripheral sitting on an I2C bus when we would enable BLE connections. The peripheral (LSM6DS3) has an internal pointer which is self updated to stream 6 channel of data to a single 2 byte buffer (so, 12 bytes for the whole 6 channels). Our firmware would repeatedly issue reads form that register (we exploited its FIFO) and keep track of how many bytes were read to correctly reconstruct the 6 signals (since only 2 bytes at time are available in the register).
The corruption would be a circular shift in the data of the signals: at some point, signal 1 would be stitched to signal 2 and so on as a circular shift. So, we missed 2 bytes and the whole data sequence would be offset.

### Debug
We checked our firmware up to reaching the single atomic I2C read. The shift would be due to a "missing" read which resulted in the cpu pointer being stale while the peripheral pointer would advance, since it had already issued the bytes to its register. This is the shift introduction.
We noted that the nrf52840 has an abstraction for the BLE stack (SoftDevice) which has the highest interrupt preemption priority, so its routine is serviced immediately. We thought that our heavy use of BLE radio in our firmware would eventually preempt the cpu while an I2C was issued and that servicing the routine would broke that read.
But the nrf52840 also have a DMA (EasyDMA), so why our cpu preemption was a problem in the first place?

### Findings
The current implementation of Wire.h for the nrf52840 does not use TWIM hardware SHORTS for repeated-START transactions, so the CPU must manually bridge the TX→RX transition each time. Enabling the SoftDevice stack installs high-priority interrupts that can preempt at any point, including between the TX and RX phases of a write-then-read I2C transaction. So, the preemption would eventually happens while the cpu would need to issue a START, but this breaks due to the BLE routine.

### Solution
The EasyDMA in the nrf52840 allows for repeated START-STOP so to not involve the cpu at all. We implemented `writeThenRead()` in this fix to do that.

### Tests

- [x] verified `writeThenRead()` with a sensor that requires repeated START (IMU LSM6DS3)
- [x] standard `beginTransmission()`/`endTransmission()`/`requestFrom()` flow still works                                                 
- [x] stress-test with BLE active confirm no START-STOP-START regression

Open to feedback on the API design, naming conventions, error handling approach or else.

### DIFF
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -33,6 +33,7 @@
 class TwoWire : public Stream
 {
   public:
+    uint8_t writeThenRead(uint8_t address, const uint8_t *write_buffer, size_t write_len, uint8_t *read_buffer, size_t read_len);
     TwoWire(NRF_TWIM_Type * p_twim, NRF_TWIS_Type * p_twis, IRQn_Type IRQn, uint8_t pinSDA, uint8_t pinSCL);
     void begin();
     void begin(uint8_t);
--- a/libraries/Wire/Wire.h
+++ b/libraries/Wire/Wire.h
@@ -32,9 +32,8 @@
 
 class TwoWire : public Stream
 {
-  public:
-    uint8_t writeThenRead(uint8_t address, const uint8_t *write_buffer, size_t write_len, uint8_t *read_buffer, size_t read_len);
-    TwoWire(NRF_TWIM_Type * p_twim, NRF_TWIS_Type * p_twis, IRQn_Type IRQn, uint8_t pinSDA, uint8_t pinSCL);
+public:
+    TwoWire(NRF_TWIM_Type *p_twim, NRF_TWIS_Type *p_twis, IRQn_Type IRQn, uint8_t pinSDA, uint8_t pinSCL);
     void begin();
     void begin(uint8_t);
     void end();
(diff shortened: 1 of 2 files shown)


fedemengo:
@hathach gentle ping for feedback 🙏

fedemengo:
@hathach gentle ping for feedback 🙏

PR #874, opened 2026-09-02
Add non-blocking timeout option to `getHvnPacket()` / `notify()`

usernameLoophole:
`BLEConnection::getHvnPacket()` always blocks up to `BLE_GENERIC_TIMEOUT` (100ms) waiting for a free HVN TX credit. If the connection is torn down while a caller is parked in that wait, the disconnect handler (`bluefruit.cpp`, `BLE_GAP_EVT_DISCONNECTED`) deletes the `BLEConnection` — and its _hvn_sem — out from under the still-blocked task. Deleting a FreeRTOS semaphore a task is blocked on is undefined behavior; on hardware this reproduces as a permanent hang.

Adds an optional `timeout_ms`/`hvn_timeout_ms` parameter to `BLEConnection::getHvnPacket()` and `BLECharacteristic::notify(conn_hdl, data, len, ...)`, defaulting to the existing `BLE_GENERIC_TIMEOUT` — every current call site is unaffected. A caller that wants to avoid the block entirely (and the race it enables) can now pass 0 for a non-blocking credit check.

Consumed by stealth30110/startDownFw#38 — see that PR's description (and docs `BLESinkWriteBug.md` in that repo) for the full investigation and hardware repro.

### DIFF
--- a/.github/workflows/platformio-package-ci.yml
+++ b/.github/workflows/platformio-package-ci.yml
@@ -1,0 +1,84 @@
+name: PlatformIO Package CI
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      base_version:
+        description: Base package version in MAJOR.MINOR form
+        required: false
+        default: "1.107"
+        type: string
+
+permissions:
+  contents: write
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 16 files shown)


usernameLoophole:
Opened against the wrong repo by mistake - this was meant for our own fork, not upstream. Closing.

